PR #34, opened 2019-09-09
Systemd Offline Upgrade Support

mmstick:


mmstick:
As soon as the most recent commit builds, this is ready for testing of both the CLI and GTK frontend.

mmstick:
- Desktop notifications are now working.
- LTS may dismiss the notifications in the info overview.

brs17:
@mmstick How do you think the following error should be handled?
```
Sep 16 12:23:40 system76-pc pop-upgrade[1127]: E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
Sep 16 12:23:40 system76-pc pop-upgrade[1127]: E: Unable to lock directory /var/lib/apt/lists/
```

Currently the release upgrade is aborted. Does this make the most sense? Would we want a different behavior in the GTK app?

mmstick:
It should wait until the lock file no longer exists, and then proceed.

mmstick:
The solution is to probe the contents of procfs for a process that owns a file descriptor which points to `/var/lib/apt/lists/lock`. I'll have a commit soon which will wait to issue apt commands until no process holds a file descriptor to the lock file.

leviport:
After an upgrade on a dual-booted machine (Pop 18.04 and Ubuntu 18.04), sources.list still says Bionic:

```
# deb cdrom:[Pop_OS 18.04 _Bionic Beaver_ - Release amd64 (20190806)]/ bionic main restricted
deb http://us.archive.ubuntu.com/ubuntu/ bionic main restricted universe multiverse
deb-src http://us.archive.ubuntu.com/ubuntu/ bionic main restricted universe multiverse
deb http://us.archive.ubuntu.com/ubuntu/ bionic-security main restricted universe multiverse
deb-src http://us.archive.ubuntu.com/ubuntu/ bionic-security main restricted universe multiverse
deb http://us.archive.ubuntu.com/ubuntu/ bionic-updates main restricted universe multiverse
deb-src http://us.archive.ubuntu.com/ubuntu/ bionic-updates main restricted universe multiverse
deb http://us.archive.ubuntu.com/ubuntu/ bionic-backports main restricted universe multiverse
deb-src http://us.archive.ubuntu.com/ubuntu/ bionic-backports main restricted universe multiverse
deb http://apt.pop-os.org/proprietary bionic main
# deb-src http://apt.pop-os.org/proprietary bionic main

deb http://apt.pop-os.org/staging/upgrade bionic main
deb-src http://apt.pop-os.org/staging/upgrade bionic main

deb http://apt.pop-os.org/proprietary disco main
# deb-src http://apt.pop-os.org/proprietary disco main
```

mmstick:
This wasn't a `.save` file? I know that I keep a backup of the original at `list.save`.

brs17:
Upon completion of the upgrade, the boot menu in the BIOS should be updated to say Pop!_OS 19.04.

This is what it looks like on laptops with a dualboot.
IMG_20190917_114136

mmstick:
As for the apt sources reverting to bionic, there is logic in the daemon that resets the apt lists if it determines that an upgrade failed to complete / was aborted.

mmstick:
The upgrade script calls this before rebooting, which should be setting it based on the `PRETTY_NAME`:

```sh
    if test -d '/sys/firmware/efi/'; then
        current_bootnum=$(efibootmgr | grep BootCurrent | awk -F' ' '{print $2}')
        new_label=$(cat /etc/os-release | grep PRETTY | awk -F'"' '{print $2}')
        efibootmgr -b "${current_bootnum}" -B
        efibootmgr -c -L "${new_label}"
    fi
```

brs17:
> The upgrade script calls this before rebooting, which should be setting it based on the `PRETTY_NAME`:
> 
> ```shell
>     if test -d '/sys/firmware/efi/'; then
>         current_bootnum=$(efibootmgr | grep BootCurrent | awk -F' ' '{print $2}')
>         new_label=$(cat /etc/os-release | grep PRETTY | awk -F'"' '{print $2}')
>         efibootmgr -b "${current_bootnum}" -B
>         efibootmgr -c -L "${new_label}"
>     fi
> ```

Well if this is not working, could it be because the `/etc/os-release` file has not yet been updated to the new version?

mmstick:
The cause of disco getting reverted back to bionic is because the pop-upgrade daemon is being started as it is upgraded to disco, which then detects that the upgrade did not complete, and proceeds to reset the apt sources back to bionic, thinking that the upgrade failed / was aborted.

mmstick:
> Well if this is not working, could it be because the `/etc/os-release` file has not yet been updated to the new version?

That's possible. The script won't exit unless that section is called, or the kernel / systemd kills the script.

brs17:
> The solution is to probe the contents of procfs for a process that owns a file descriptor which points to `/var/lib/apt/lists/lock`. I'll have a commit soon which will wait to issue apt commands until no process holds a file descriptor to the lock file.

Let's leave some log or mention of why `pop-upgrade` is just sitting there. Other than that, the fix for this looks good!

leviport:
Plymouth looks weirdly sized on the troublesome monitor in the lab: 

IMG_20190918_093218

The screenshot also shows something I saw a lot through this upgrade, which is the text saying "Unpacking Unpacking"

mmstick:
That seems to be an older version of `pop-upgrade`. The build servers have run out of disk space when building for bionic.

mmstick:
The displacement could also be that the system is reporting that the screen is running at a higher resolution than what is actually being displayed.

brs17:
> Upon completion of the upgrade, the boot menu in the BIOS should be updated to say Pop!_OS 19.04.
> 
> This is what it looks like on laptops with a dualboot.
> IMG_20190917_114136

Now, after an upgrade I see this: 

bootmenu

NOTE: This is a different install so there shouldn't be a Windows Boot Manager, but before doing the upgrade it said 18.04 like in the quoted image.

brs17:
On a thelio-major-b2, the following is the BIOS boot menu on a system dualbooting Ubuntu 18.04 and Pop!_OS 19.04 (after upgrading from 18.04).

Here it looks like 19.04 is updated correctly.



However, when selected, the Pop!_OS 19.04 option does not boot. The `UEFI OS` option had to be selected to get Pop!_OS to boot.

mmstick:
Seems I also need to specify the bootloader file path. Using the `-c` argument isn't enough.

leviport:
We saw a failed upgrade happen on the Darter this morning. We saw this output:
Screenshot from 2019-09-23 08-42-30
Screenshot from 2019-09-23 08-44-44
Sources.list still showed everything was set to Bionic, which is expected, and all packages were up to date (it was run with updates available). Re-running `pop-upgrade release upgrade systemd` seemed to work correctly.

mmstick:
Those screenshots appear to come from an older version of pop-upgrade. Commit `0e3bbf8` changed the messages in the CLI client so that `Dpkg: ` and `Release Event:` are no longer written to the logs.

mmstick:
If you get `failed to unpack` warnings, you may be running the version of `pop-upgrade` in the `master` branch, rather than the `upgrade` branch.

leviport:
I think you're right: the Plymouth screen also looked like an old version. Sorry about that.

brs17:
@mmstick I removed network access during `pop-upgrade release upgrade systemd` and got this:

```
Progress: 99%
Setting up system76-driver
Setting up mutter
Setting up system76-driver-nvidia
Setting up gnome-shell
Processing triggers for linux-image-5.0.0-30-generic
Processing triggers for linux-image-5.3.0-12-generic
Processing triggers for libc-bin
Event: waiting on a process holding the apt lock files
Event: ensuring that system-critical packages are installed
Event: updating the source lists to the new release
Incompatible repositories detected:
    Error: http://us.archive.ubuntu.com/ubuntu/:
        failed to lookup address information: Name or service not known
    Error: http://apt.pop-os.org/proprietary:
        failed to lookup address information: Name or service not known
    Error: http://apt.pop-os.org/staging/upgrade:
        failed to lookup address information: Name or service not known
    Error: http://ppa.launchpad.net/system76/pop/ubuntu:
        failed to lookup address information: Name or service not known
Requesting user input:
[INFO ] pop_upgrade src/cli/mod.rs:399: sending message
    Keep repository: (http://us.archive.ubuntu.com/ubuntu/)? y/N: 
```

I probably shouldn't be prompted to remove these ppas.

brs17:
Just attempted to unplug Ethernet (disconnecting network) at 99% of running `pop-upgrade release upgrade systemd` and got this:

```
Progress: 92%
Processing triggers for libc-bin
Processing triggers for udev
Processing triggers for systemd
Processing triggers for man-db
Processing triggers for gnome-menus
Processing triggers for dbus
Processing triggers for hicolor-icon-theme
Processing triggers for mime-support
Processing triggers for ureadahead
Processing triggers for desktop-file-utils
Processing triggers for install-info
Processing triggers for libglib2.0-0:amd64
Progress: 93%
Progress: 94%
Setting up gnome-control-center
Setting up libmutter-2-0:amd64
Progress: 95%
Setting up gir1.2-mutter-2:amd64
Progress: 96%
Setting up mutter
Progress: 97%
Setting up gnome-shell
Progress: 98%
Processing triggers for initramfs-tools
Processing triggers for linux-image-5.0.0-30-generic
Processing triggers for linux-image-5.3.0-12-generic
Progress: 99%
Processing triggers for libc-bin
pop-upgrade: dbus client error: calling Status method failed
system76@pop-os:~$ pop-upgrade release upgrade systemd
pop-upgrade: dbus client error: calling ReleaseCheck method failed
```

Now I am unable to re-run `pop-upgrade release upgrade systemd` getting `pop-upgrade: dbus client error: calling ReleaseCheck method failed` every time.

brs17:
> Sounds like the daemon stopped

@mmstick do you have any ideas how to prevent ending up in a situation like this? Especially when the system is just downloading packages for the upgrade, I would imagine users could be doing anything and everything under the sun so this scenario very well could happen again.

mmstick:
Depends on what caused it to stop. It's possible that systemd sent a signal to stop it. I could enable signal handlers to listen for the signals that try to stop or terminate the process.

leviport:
After modifying `/etc/vim/vimrc` and performing an upgrade, my changes to vimrc were preserved, but I did not receive a prompt about keeping my version or replacing it with the maintainer's version. This seems correct, but I believe a `do-release-upgrade` asks that question, so I just wanted to make sure that was intended behavior for pop-upgrade.

jackpot51:
@mmstick could you explain how debconf questions are being handled?

mmstick:
All apt commands are executed with `DEBIAN_FRONTEND` set to `noninteractive`, which makes debconf answer the questions without prompting the user.

leviport:
Upgrades seem to be releasing the lock file multiple times before the reboot. I see "Updates Available" notifications, and pop-upgrade says it's waiting on the lock file before it proceeds. I can also run `sudo apt upgrade` while it's fetching new packages, which feels weird.

leviport:
Another thing we may want to consider: currently when the offline upgrade is being performed and the machine shuts down, booting it back up takes it back to the Plymouth screen, saying that a power failure was detected and it is retrying the upgrade. This is awesome. However another power failure, as unlikely as that may be in the real world, takes us back to Plymouth on the next boot, saying that the upgrade failed and it is bailing. It then returns to GDM and allows you to sign in to what looks like 18.04. The sources.list still has Disco entries though, and apt is left in a pretty broken state. 

My thoughts are that we may want to either bail more gracefully and restore sources.list to the previous version, or to continue attempting the upgrade either until it succeeds.

mmstick:
Unfortunately, we can't hold a file descriptor to the lock file exclusively throughout the entire process because we wouldn't be able to use `apt` until we release it, so there's always a chance that another process was waiting for its turn to use `apt`. We could, however, patch the appcenter to look for our upgrade file, and disable update checks when found, and I can hold the file while fetching packages.

The desktop notification happens because the appcenter looks for `/var/run/reboot-required`, and creates the notification when found. We could also patch it to ignore that file if a release upgrade is in progress.

As for restoring the sources list on three failed attempts to upgrade, this won't reverse the packages that were installed, and it may not be possible to downgrade them in a half-installed state. So the only options are to drop the user into the half-updated system for them to manually repair, or to continually retry indefinitely.

mmstick:
There's now an `upgrade` branch of the shop which prevents the restart notification from appearing if a release upgrade is in progress.

leviport:
After a (forced) upgrade from 19.04 to 19.10 on serw11, there were some dependency issues with the Nvidia driver and I could not install updates with `sudo apt update && sudo apt upgrade`. The resulting output is attached.

[1904-to-1910-error.txt](https://github.com/pop-os/upgrade/files/3691595/1904-to-1910-error.txt)

mmstick:
@leviport This should be fixed by https://github.com/pop-os/nvidia-graphics-drivers/pull/27

leviport:
After using pop-upgrade to go from 19.04 with dark themes enabled, to 19.10, it looks like the default theme is set to Adwaita upon first boot. After opening the "appearances" settings, the theme immediately switches back to pop-slim-dark. The setting appears to stick after visiting "appearances".

[dark-theme-bug.zip](https://github.com/pop-os/upgrade/files/3700089/dark-theme-bug.zip)

mmstick:
@leviport Sounds like a possible kernelstub issue. It should detect 5.3 as the latest kernel, rather than 5.0.

leviport:
After a little more digging WRT that kernel version thing, I noticed fresh 19.04 installs upgrade properly to 5.3. It's only installs that start at 18.04 that seem to get stuck on the older kernel.

mmstick:
I wonder if that's because they still have the 4.8 kernel installed?

leviport:
Hmm, I'm only seeing 4.15 installed on any I've run upgrades on. Even so, they're set to 5.0.0-23 fresh off the imaging server. You think having 4.15 there would break it?

mmstick:
Having the 4.15 kernel originally installed would be the main difference in kernels between 18.04 and 19.04. I'm not sure why that would make a difference here, though.

leviport:
Kernel issue was addressed via https://github.com/pop-os/kernelstub/pull/35

leviport:
Non-blocking, but currently when I'm prompted whether or not I want to keep a broken PPA, it prompts me to answer (y/N). It works as expected, but if I enter something else (the number 1, for example), I get the message, "Could not parse 1 as bool". I know this will likely be hidden behind the GTK interface in the future, but maybe this could be a friendlier message?

Also fun: If I enter "true" or "false" they also work as Y or N, respectively. This probably isn't an issue, just something I noticed.

mmstick:
What should we display when the input is invalid?

mmstick:
I could have the prompt reply that it only accepts either `y` or `n`.

brs17:
> Non-blocking, but currently when I'm prompted whether or not I want to keep a broken PPA, it prompts me to answer (y/N). It works as expected, but if I enter something else (the number 1, for example), I get the message, "Could not parse 1 as bool". I know this will likely be hidden behind the GTK interface in the future, but maybe this could be a friendlier message?
> 
> Also fun: If I enter "true" or "false" they also work as Y or N, respectively. This probably isn't an issue, just something I noticed.

@mmstick I'm not terribly concerned about this, and honestly it is relatively straight forward so I would say don't worry about this.

mmstick:
Changing the retpoline gcc cflags in the Makefiles of the kernels prior to 5.3 allows dkms to successfully build the nvidia dkms module + system76 dkms kernel modules. Canonical may fix this by releasing updated kernels for 18.04 and 19.04, but we can fix this by [automatically applying the changes needed](https://bugs.launchpad.net/ubuntu/+source/gcc-9/+bug/1830961/comments/4).

jackpot51:
I will have to think about it. We have more than just those two - we also have system76-acpi-dkms and system76-io-dkms

mmstick:
system76-acpi-dkms and system76-io-dkms also successfully build with the change

jackpot51:
Could this be added to the dkms package instead?

jackpot51:
We would have to edit those 4 dkms packages as well as the virtualbox-dkms package, and that would take quite a while to run through QA

mmstick:
We could apply the fix either here or in the dkms package, and we can close the PR for NVIDIA.

jackpot51:
I see, whichever requires the least change

mmstick:
@pop-os/quality-assurance Calling `pop-upgrade release upgrade systemd -f` on 18.04 will upgrade it directly to 19.10. It essentially picks the current development branch, so once we've published release ISOs for 19.10, it will start pointing to 20.04.

mmstick:
I just successfully performed a release upgrade directly from 18.04 to 19.04, with zero DKMS errors. All packages upgraded successfully.

jackpot51:
How does it do to 19.10? That was where the issues happened

mmstick:
When it executes the `repair()` method before initiating the upgrade, it applies the dkms gcc9 fix in c929296 to the makefiles for kernel sources older than 5.3, so the upgrade succeeds without error.

leviport:
When attempting to do an 18.04 -> 19.04 -> 19.10 upgrade on the Meerkat, I'm seeing the following error: 

```
system76@pop-os:~$ pop-upgrade release upgrade systemd -f
/sys/firmware/dmi/tables/smbios_entry_point: Permission denied
/dev/mem: Permission denied
Release upgrade status: release upgrade aborted: failed to apply system repair before upgrade: unable to apply dkms gcc9 fix: No such file or directory (os error 2)
```

This is after the 18.04 -> 19.04 which worked fine.

mmstick:
Does the `/lib/modules` directory exist on that system? And does each directory in that directory have a `build/Makefiles` path?

mmstick:
The latest commit will check for either `/lib/modules` or `/usr/lib/modules`, and only consider kernels which have an existing path to their makefile.

leviport:
You're fast! In case the info helps, I'll post it anyway: 

There were 3 directories in `/lib/modules` , but it looks like one doesn't have a `build` directory:

```
system76@pop-os:/lib/modules$ ls
5.0.0-21-generic  5.0.0-23-generic  5.0.0-31-generic
system76@pop-os:/lib/modules$ ls 5.0.0-31-generic/
build   modules.alias      modules.builtin.bin  modules.devname  modules.symbols      vdso
initrd  modules.alias.bin  modules.dep          modules.order    modules.symbols.bin
kernel  modules.builtin    modules.dep.bin      modules.softdep  updates
system76@pop-os:/lib/modules$ ls 5.0.0-23-generic/
build   modules.alias      modules.builtin.bin  modules.devname  modules.symbols      vdso
initrd  modules.alias.bin  modules.dep          modules.order    modules.symbols.bin
kernel  modules.builtin    modules.dep.bin      modules.softdep  updates
system76@pop-os:/lib/modules$ ls 5.0.0-21-generic/
modules.alias      modules.builtin.bin  modules.dep.bin  modules.softdep  modules.symbols.bin
modules.alias.bin  modules.dep          modules.devname  modules.symbols
```

mmstick:
I see that the `21-generic` kernel there doesn't have a `build` symlink. It must be that kernels that are removed will have their source symlink removed, so the latest commit should fix this.

leviport:
After an upgrade to 19.04 from 18.04, I routinely see this message when running `sudo apt upgrade` for the first time after the pop-upgrade:

```
1 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
3 not fully installed or removed.
```

I found out which packages they were with `dpkg -C`:

```
system76@pop-os:~$ sudo dpkg -C
The following packages have been unpacked but not yet configured.
They must be configured using dpkg --configure or the configure
menu option in dselect for them to work:
 acpi-support         scripts for handling many ACPI events
 pop-desktop          Pop Desktop Metapackage

The following packages are only half configured, probably due to problems
configuring them the first time.  The configuration should be retried using
dpkg --configure <package> or the configure menu option in dselect:
 acpid                Advanced Configuration and Power Interface event daemon

system76@pop-os:~$ apt policy pop-desktop acpi-support acpid
pop-desktop:
  Installed: 1.2.12~1556292703~19.04~b6b743d
  Candidate: 1.2.12~1556292703~19.04~b6b743d
  Version table:
 *** 1.2.12~1556292703~19.04~b6b743d 1002
       1002 http://apt.pop-os.org/staging/master disco/main amd64 Packages
       1002 http://apt.pop-os.org/staging/master disco/main i386 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu disco/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu disco/main i386 Packages
        100 /var/lib/dpkg/status
acpi-support:
  Installed: 0.143
  Candidate: 0.143
  Version table:
 *** 0.143 500
        500 http://us.archive.ubuntu.com/ubuntu disco/main amd64 Packages
        100 /var/lib/dpkg/status
acpid:
  Installed: 1:2.0.31-1ubuntu2
  Candidate: 1:2.0.31-1ubuntu2
  Version table:
 *** 1:2.0.31-1ubuntu2 500
        500 http://us.archive.ubuntu.com/ubuntu disco/main amd64 Packages
        100 /var/lib/dpkg/status
```

After the `apt upgrade`, everything is fine, but I wanted to point this out in case it was important.

mmstick:
I've just installed 18.04 NVIDIA, upgraded directly to 19.10, and verified that no packages required setting up, and that Linux 5.3 was installed and running.

mmstick:
HWE kernels are now removed after a successful upgrade

leviport:
It looks like pop-upgrades from 19.04 to 19.10 do not work unless the system is first fully updated with `sudo apt update && sudo apt upgrade`. Attached are console output and journalctl logs for a `pop-upgrade release upgrade systemd -f` performed on a fresh 19.04 install that didn't have software updates applied first.

[darp5-1904-to-1910-journalctl.txt](https://github.com/pop-os/upgrade/files/3731710/darp5-1904-to-1910-journalctl.txt)
[darp5-1904-to-1910-terminal.txt](https://github.com/pop-os/upgrade/files/3731711/darp5-1904-to-1910-terminal.txt)

mmstick:
`failed to rename "/var/cache/apt/archives/partial/linux-headers-5.0.0-31_5.0.0-31.33_all.deb" to "/var/cache/apt/archives/linux-headers-5.0.0-31_5.0.0-31.33_all.deb" }))`

Does the source file exist?

data/pop-upgrade.sh
```diff
@@ -1,8 +1,13 @@
-#!/bin/sh
+#!/bin/bash
```
jackpot51: There are loads of unquoted variables here. I think this can be taken care of post-merge

src/daemon/sighandler.rs
```diff
@@ -0,0 +1,61 @@
...
+    extern "C" fn handler(signal: i32) {
+        let signal = match signal {
+            libc::SIGINT => Signal::Interrupt,
+            libc::SIGHUP => Signal::Hangup,
+            libc::SIGTERM => Signal::Terminate,
+            libc::SIGTSTP => Signal::TermStop,
+            _ => unreachable!(),
+        };
+
+        warn!("caught {} signal", signal);
```
jackpot51: You should not use global variables in signal handlers. If the logger is locked when the signal occurs, this could cause a deadlock. The atomic on the next line is okay because it has known ordering.

PR #46, opened 2019-10-18
Upgrade GTK Widget (continued)

mmstick:
Includes the last remaining changes before we release the integration of the GTK widget in GNOME Control Center. The previous PR had to be merged so that our CI would build GCC with the new C headers.

### DIFF
--- a/gtk/src/widgets/dialogs/upgrade.rs
+++ b/gtk/src/widgets/dialogs/upgrade.rs
@@ -44,7 +44,7 @@ impl UpgradeDialog {
             "distributor-logo",
             "Upgrade",
             "Reboot & Upgrade",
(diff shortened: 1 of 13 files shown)


gtk/src/widgets/dialogs/upgrade.rs
```diff
@@ -44,7 +44,7 @@ impl UpgradeDialog {
             "distributor-logo",
             "Upgrade",
             "Reboot & Upgrade",
-            &gtk::STYLE_CLASS_SUGGESTED_ACTION,
+            &gtk::STYLE_CLASS_DESTRUCTIVE_ACTION,
```
isantop: This appears to be correct!

mmstick:
I just amended the last commit so that the label is correctly set when a reboot is already scheduled.

PR #54, opened 2019-10-28
improvement(daemon): Faster snapd pre-depends search

mmstick:
The snapd pre-depends lookups are currently really slow due to spawning an instance of apt-cache for each dependency found of snapd. This update moves the pre-depends logic to the apt-cli-wrappers crate, which does a single call to apt-cache to fetch the pre-depends packages.


src/release/mod.rs
```diff
@@ -390,7 +390,15 @@ impl<'a> DaemonRuntime<'a> {
         // notifications.
         const APPCENTER: &str = "io.elementary.appcenter";
 
-        for proc in procfs::all_processes() {
+        let processes = match procfs::process::all_processes() {
+            Ok(proc) => proc,
+            Err(why) => {
+                warn!("failed to running processes: {}", why);
```
crawfxrd: This missing a word?

mmstick: Indeed

Issue #55, opened 2019-11-07
Write/Publish changelogs for all releases between 18.04 and 19.10.


brs17:
The changelog for 18.10 and 19.04 need to be created to be utilized in displaying the changelog from 18.04->19.10.

mmstick:
Adding an 18.10 file here will make that happen: https://github.com/pop-os/upgrade/tree/master/changelogs

PR #58, opened 2019-11-12
Add changelogs for 19.04 and 18.10

brs17:
I created changelogs based on the content provided in System76 blog posts from release.

https://blog.system76.com/post/184281497363/popos-1904-is-here

https://blog.system76.com/post/179217201328/see-what-changes-have-been-orbiting-popos

This pr resolves https://github.com/pop-os/upgrade/issues/55

### DIFF
--- a/changelogs/18.10
+++ b/changelogs/18.10
@@ -1,0 +1,17 @@
+* Pop!_OS Shop Application previews now load faster.
+
+* Improvements to the Pop!_Shop UI to prevent it from freezing.
+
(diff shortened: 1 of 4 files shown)


brs17:
I noticed while putting this together the fact that the 19.10 release notes probably need to be updated to say that the slim mode was removed since the 19.04 changelog mentions the addition of slim mode.

mmstick:
I've added the new changelogs to the src/changelogs module, so they should be appearing in the dialog now.

PR #61, opened 2019-11-19
Upgrade recovery partition on release upgrade

### LINKED ISSUE #60: Add updating recovery partition to systemd upgrade process
When upgrading to a new OS release, the recovery partition should be updated as well (without having to select a specific `update recovery partition` button).

mmstick:
Closes #60


README.md
```diff
@@ -13,99 +29,6 @@ following details:
...
-- `InstallingPackages` (`4`): ensuring that system-critical packages are isntalled
-- `UpdatingSourceLists` (`5`): updating the source lists to the new release
-- `FetchingPackagesForNewRelease` (`6`): fetching packages for the new release
-- `AttemptingLiveUpgrade` (`7`): attempting live upgrade to the new release
-- `AttemptingSystemdUnit` (`8`): creating a systemd unit for installing the new release
-- `AttemptingRecovery` (`9`): setting up the recovery partition to install the new release
-- `Success` (`10`): new release is ready to install
-- `SuccessLive` (`11`): new release was successfully installed
-- `Failure` (`12`): an error occurred while setting up the upgrade
-
```
brs17: You could probably put this in a HACKING.md/CONTRIBUTING.md if you wanted.

leviport:
This is looking good so far, but I did notice something that we might want to consider. If the user initiates the upgrade and then cancels it, the changes revert and the upgrade can be done at a later time. With this patch, the recovery partition is not reverted to its previous version when the upgrade is cancelled.

I think this is what we want anyway since there aren't really negatives to having an updated recovery partition, but it makes me wonder if cancelling the upgrade should revert everything, including the recovery partition. If the user puts off the upgrade for too long, it does keep the machine in a strange state where the recovery partition is a newer version than the installed OS. Thoughts?

I have a couple more things I want to test, but I'm almost ready to approve it as long as the above behavior is what we want.

brs17:
> This is looking good so far, but I did notice something that we might want to consider. If the user initiates the upgrade and then cancels it, the changes revert and the upgrade can be done at a later time. With this patch, the recovery partition is not reverted to its previous version when the upgrade is cancelled.
> 
> I think this is what we want anyway since there aren't really negatives to having an updated recovery partition, but it makes me wonder if cancelling the upgrade should revert everything, including the recovery partition. If the user puts off the upgrade for too long, it does keep the machine in a strange state where the recovery partition is a newer version than the installed OS. Thoughts?
> 
> I have a couple more things I want to test, but I'm almost ready to approve it as long as the above behavior is what we want.

I think we should upgrade the recovery partition after the pop-up as if you cancel the upgrade, you now have a recovery partition with a newer OS version than the current system. This is not good and should not be expected behavior.

@mmstick is it possible to do the upgrade during the upgrade process as part of the systemd service?

mmstick:
Technically everything is happening in the systemd service. The GTK widget just sends signals to the daemon to tell it what to do, and then listens for events

brs17:
Closing since implementation details are still missing and this needs to be implemented after https://github.com/pop-os/gnome-control-center/issues/67, https://github.com/pop-os/gnome-control-center/issues/68 and other stuff mentioned here https://github.com/pop-os/upgrade/wiki/Designs-for-ther-OS-Upgrade-&-Refresh

PR #66, opened 2019-11-25
Improved OS Notifications Design

mmstick:
This implements the improved design for OS upgrades


mmstick:
This is the first part of a series of PRs that will improve the design of our upgrade panel. The second PR will build upon this in order to add support for the recovery partition.

mmstick:
I'm removing the last commit which added the OS info widget.

mmstick:
The dismiss button isn't being added to the size group

Issue #32, opened 2019-01-04
Pop!_OS Dist Upgrades


WatchMkr:
Offline dist-upgrades.

* Notify user that a OS version update is available to download
* Require user action to start download (Setting > About)
* Notify user that they can start the upgrade once the package download is complete
* Download additional packages if things changed between when the packages were first downloaded and the user decided to start the upgrade.
* Upgrade UI elements in Settings > About (notification takes the user here)
* Reboot into the recovery partition and installer
* Use the progress window in the installer to show upgrade progress
* Automatically reboot once complete

Settings > About > New version available to download

![pop about settings](https://user-images.githubusercontent.com/3430283/51066611-81647e00-15c8-11e9-84c4-df9ec33e1d8f.png)

Download progress

![pop upgrade downloading](https://user-images.githubusercontent.com/3430283/51066636-9d681f80-15c8-11e9-947a-cb94479c602f.png)

Download complete

![pop upgrade download complete](https://user-images.githubusercontent.com/3430283/51066650-ab1da500-15c8-11e9-96cb-1007ba82e1e1.png)

Confirm upgrade window

![pop upgrade confirmation](https://user-images.githubusercontent.com/3430283/51066654-b1ac1c80-15c8-11e9-8ebb-e7592f473a64.png)

Running the current version

![pop upgrade current version](https://user-images.githubusercontent.com/3430283/51066657-b83a9400-15c8-11e9-9e4d-3b6fa5984ee7.png)

brs17:
With pop-upgrade out on 18.04, 19.04, and 19.10 I think this issue can be closed. A new issue should be created to discuss adding additional features to the `About` page (Refresh install, etc).

PR #67, opened 2019-11-25
Most recent designs

maria-komarova:
There are three features: OS Upgrade, Recovery Partition Update and OS Refresh.

There are also three major use cases. 
1. Someone has the OS up to date. In this case the screen would include the version of the OS, the message that the system is up to date and the Refresh OS option. When there is a major update to the Recovery Partition that comes before the OS upgrade the Recovery Partition (RP) should appear underneath the OS Upgrade Information but above the Refresh OS.

2. Someone needs to Refresh OS when the RP update was not installed. Once the RP update is available, we notified the user but the user ignored the notification or didn't have time. The user initiates the Refresh OS, the system finds that the feature in the RP is missing and requires the RP update. So instead of the pop up to confirm the reboot, the user gets a pop-up with the message that the system requires updating the RP and downloading X GB. Once the user confirms, the system starts download and then install of the RP. The user gets feedback through the progress bar in the pop-up. It would be nice to change the message on top of the progress bar from downloading to updating to provide more feedback. Once the RP is updated the user is prompted to reboot. 
It is possible that the user's internet connection would be bad so there is an error message in the mockups.

3. A user is on LTS. So there is the option to dismiss notifications. It is the only difference.


brs17:
On RefreshOS-RPneded.png  - should we show a progress bar for downloading/updating the recovery partition? For such a large download it would probably be a good idea...I guess the question then is should that be in the pop-up, or should the pop-up close?

brs17:
When will we show RefreshOS-ready.png vs UpToDate-Refresh.png ? Sh/could we just use UpToDate-Refresh.png ?

brs17:
Probably assumed but just to be clear, we will want the `Reboot and Refresh` button on RfreshOS-DownloadingRP.png to be inactive.

brs17:
With the error message on RefreshOS-RPerror.png, is this error message only expected due to networking/downloading issues? Perhaps we can be smarter about this and show a network connectivity message on RefreshOS-RPneded.png ? @mmstick can correct me if I'm wrong, but I recall us doing some clever stuff in regards to dealing with downloading and unreliable networks.

It does make me wonder however, if the downloading/updating of the recovery partition should be done(downloaded and updated) on the main GNOME Control Center page and not in a popup. Maybe make it similar to RecoveryUpdating.png ?

mmstick:
We usually place error messages into a GtkInfoBar within an overlay at the top of a panel.

The buttons in the widget currently do support dynamically alternating between showing a button, and showing a progress bar, so we could also use them for recovery partition updates. We could exchange the progress bar for an updated button once it completes, along with displaying a notification that the recovery partition has been updated.

If we're also including the extraction of the ISO to the recovery partition after it completes, it's very possible for an error to occur while replacing the old recovery files with the new recovery files.

maria-komarova:
Working on designs to incorporate comments. The idea is to show but disable the Refresh option until users update the recovery partition (include requiring to update the recovery partition in the explanation under the Refresh option). That would take care of showing any progress bars in the popup and of the possibility of the error. Please, correct me if I misunderstand. That would also eliminate the need for any dynamic buttons. 
I'll adjust the wording wherever needed to avoid using too many updates etc. 

Below is the list of all the cases I can think of with what we should show in the settings accordingly (6 cases total). Please let me know if I missed something. I'll adjust the designs in the meanwhile.

0. Slim chance of this scenario but who knows:
User has the OS and the recovery partition up to date (installed today) - show OS is up to date, Refresh OS option enabled. 

1. User has system installed with the Refresh option in the recovery partition.
- 19.10 - show system 19.10, recovery partition update available, Refresh OS option enabled. 
- anything before - show system version, OS upgrade available, Recovery Partition Update available, Refresh OS option enabled.  

2. User is on any system installed before the Refresh option was incorporated into the recovery partition:
- upgraded to 19.10 - show system 19.10, recovery partition update available, Refresh OS option disabled. Enable Refresh OS option once a user updates the recovery partition. 
- anything before that - show system version, OS upgrade is available, Refresh OS option there but disabled (requires upgrading to 19.10). Enable once a user upgrades. 

3. User is on LTS (before the Refresh option was incorporated into the recovery partition):
show system version, Upgrade OS available with the ability to dismiss notifications, recovery partition update available (no ability to dismiss notifications, keep reminding because we want people to update it), Refresh OS option disabled. Enable Refresh OS option once a user updates the recovery partition.

brs17:
Discussed in person, OldOS-Upgrade-Refresh.png should not be a possible scenario because the user should always have the ability to refresh os (to fix a broken system).

maria-komarova:
Updated the designs according to the in-person conversation.

WatchMkr:
For Refresh OS, let the user knows what's going to happen when they click the button. "Reboot & Refresh" rather than just "Refresh". This confirmation looks different than the OS Upgrade and Firmware installation confirmation dialogs. They should be consistent.

If there's going to be a separate Recovery Partition Update section, it should always be present and tell the user that the recovery partition is up-to-date when there are no updates available.

maria-komarova:
I looked at the Upgrade from 18.04 to 19.10 on the virtual machine to keep the confirmation the same. Is there somewhere else I can look? 
The button on Upgrade also says Upgrade and once the confirmation pops up the button says "Reboot & Upgrade". At least that's what I see on the virtual machine.

brs17:
> I looked at the Upgrade from 18.04 to 19.10 on the virtual machine to keep the confirmation the same. Is there somewhere else I can look?
> The button on Upgrade also says Upgrade and once the confirmation pops up the button says "Reboot & Upgrade". At least that's what I see on the virtual machine.

That is the best place to look at what the upgrade looks like. We have slightly out of date (not matching what we released) here: https://github.com/pop-os/upgrade/issues/32

maria-komarova:
Closing this pull request because I opened a new issue with updated designs: [https://github.com/pop-os/upgrade/issues/69]
Just searching for the best ways for us to share the designs and discuss them.

PR #70, opened 2019-12-02
Ability to cancel recovery upgrades

mmstick:
This is required for implementing the recovery button in the GTK widget

- Adds a Cancel method to the daemon for clients to signal the intent to cancel an operation
- Added support for cancelling a recovery upgrade
- Added a cancel subcommand to the CLI frontend
- The signal handler is now activated for the CLI frontend, to call the Cancel method, instead of terminating


mmstick:
Rebasing this on master, since it doesn't depend on the notification PR

PR #71, opened 2019-12-03
Add Refresh OS Section

mmstick:
Does not add support for detecting if the recovery partition needs to be updated. For that, we will need to have [deckard implement a feature for fetching the last urgent build number & date](https://github.com/pop-os/deckard/issues/7).

### DIFF
--- a/gtk/src/events/mod.rs
+++ b/gtk/src/events/mod.rs
@@ -246,10 +246,15 @@ fn connect_upgrade(state: &mut State, widgets: &EventWidgets, is_lts: bool, rebo
         Ok(eol) => {
             let (y, m, d) = eol.ymd;
             match eol.status() {
-                EolStatus::Exceeded | EolStatus::Imminent => Some(fomat!(
+                EolStatus::Exceeded => Some(fomat!(
+                    "Support for Pop!_OS " (eol.version) " has ended. "
+                    "Security and application updates are no longer provided for Pop!_OS " (eol.version) ". "
+                    "Upgrade to Pop!_OS " (eol.version.next_release()) " to keep your computer secure."
+                )),
+                EolStatus::Imminent => Some(fomat!(
                     "Support for Pop!_OS " (eol.version) " ends "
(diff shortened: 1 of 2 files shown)


mmstick:
That makes sense, yes. I will separate the recovery partition update functionality.

mmstick:
Because the recovery upgrade PR depends on changes in the refresh OS PR, I will withhold these commits in a local branch until this PR is merged.

mmstick:
@pop-os/quality-assurance This is now ready again

leviport:
reboot reefresh

I like the extra information in this window, but I think we may want to have @maria-komarova take a look at the revised verbiage since it differs a bit from the original design.

maria-komarova:
I love that we can give a user a better idea of what exactly is going to happen! I think we can make a few minor tweaks to make the information easier to process. Here are the the suggested wording changes (the UI is merely for the context, no changes there):
popup

gtk/src/lib.rs
```diff
@@ -26,55 +31,75 @@ use std::{
...
-        let upgrade = Section::new("<b>OS Upgrade</b>");
 
         let dismisser = gtk::ListBoxRow::new();
         upgrade.list.add(&dismisser);
 
         let button_sg = cascade! {
             gtk::SizeGroup::new(gtk::SizeGroupMode::Both);
-            ..add_widget(&upgrade.option.button);
+            ..add_widget(&upgrade.options[0].button);
+            ..add_widget(&recovery.options[0].button);
```
crawfxrd: Should this be using `REFRESH_OS` for the index?

mmstick: Yes, it should be explicitly declared here.

mmstick:
It seems that `bootctl set-oneshot` is unreliable. Even though it cites that the recovery partition is the default boot loader entry, systemd-boot will ignore this, and continue to boot into the default entry.

mmstick:
`bootctl set-default` works, so we will need to add a check at system startup to change the default if the default is currently set to the recovery partition.

PR #75, opened 2019-12-16
Prevent release notifications on LTS if the install time exceeds the next release time

### LINKED ISSUE #74: Dismiss upgrade notification on LTS install post non-LTS release
When an LTS release is installed after the release of a non-LTS version, the user should not be notified of a new Pop!_OS release until the next version comes out. (e.g. Install 18.04 in December of 2019 and should not be notified of a new release until 20.04).

mmstick:
Closes #74


mmstick:
It should still work after a system restart, since the change only applies to the CLI client which is responsible for creating the notifications.

brs17:
> It should still work after a system restart, since the change only applies to the CLI client which is responsible for creating the notifications.

I have rebooted a number of times with no success.

mmstick:
I will notify you when I get it working here.

mmstick:
Found issue. Error in logs: `19.10 is not a valid codename`.

mmstick:
@brs17 Seems to be working now

mmstick:
So it needs to be a once-only operation

mmstick:
@brs17 This is working now

PR #78, opened 2020-01-08
feat: Display EOL notifications one month before the EOL date

### LINKED ISSUE #77: Support for the previous OS release ends
We should inform our users that the support for 19.04 ends January 18. Hence, the updates for the OS upgrade page. We should probably also have a notification or even warning but let's start with the page in the settings. 
The change is adding another sentence under "Pop!_OS 19.10 is available". See below.
OSupgrade-oldOS-EOL (4)

mmstick:
EOL date is derived from https://github.com/pop-os/ubuntu-version/commit/d344d4f5309ec6a77276afa6decf42f26db3c601

Closes #77


src/cli/mod.rs
```diff
@@ -518,6 +526,22 @@ fn installed_after_release(next: &str) -> bool {
...
+            fomat!(
+                "Support for Pop!_OS " (current) " ends "
+                (Utc.ymd(y as i32, m, d).format("%B %d, %Y"))
+            ),
+            fomat!(
+                "This computer will soon stop receiving updates. Upgrade to Pop!_OS "
+                (next) " to keep your computer secure."
+            ),
+        ),
+        _ => ("Upgrade Available".into(), fomat!("Pop!_OS " (next) " is available to download")),
```
crawfxrd: Is displaying this valid if `eol_impending()` fails?

mmstick: Even if a failure to get the EOL status occurs, we still know that an upgrade is available.

mmstick:
One change left, and this will be ready. The sublabels in options need to be line-wrapped

mmstick:
This is good now

jackpot51:
@maria-komarova had mentioned wanting to have a post-EOL message ready.

leviport:
Looking pretty good so far! There are a couple of UI things I wanted to bring up for discussion. 

First, the amount of text in the notification causes some of it to be cut off until the user hovers their cursor over the notification.
Pre-hover:
Screenshot from 2020-01-09 11-54-45

Post-hover:
Screenshot from 2020-01-09 12-07-48

Second, it looks like the download button got taller. This may be taken as encouragement or emphasis though, so maybe we're cool with it:
Screenshot from 2020-01-09 11-55-04

mmstick:
The button defaults to being the height of the the two rows it occupies. It can be vertically center-aligned, but then it looks too small. I can look into a way to make it appear to have the default size while center-aligned.

mmstick:
Pushing a new commit shortly that will improve the EOL logic to support handling a post-EOL condition. Although since we do not have a design on that, I will have it do the same as when it is in an impending EOL status.

maria-komarova:
I'll push the new commit. Although I hope to have the designs ready not later than tomorrow (with regards to all the possible discussions).

mmstick:
We don't have control over the text being ellipsed, so we'll want to keep the body of the notification brief.

maria-komarova:
So we never had a longer notification? What did we do before in those cases?
We can cut the last sentence to "Upgrade to Pop!_OS 19.10" although I have seen GNOME designs with the notification being the way it is now, as awkward as it is. At least, I have seen it in the list of the notifications.

mmstick:
Screenshot of vertically-aligned button:

Screenshot from 2020-01-09 13-19-58

maria-komarova:
I think if we make the text of the "Support ends..." smaller, the button should look fine vertically aligned. There is a precedent for that in Settings.

mmstick:
Still wraps with a `0.9` scale. I'll see about adding a margin to the label inside the button

maria-komarova:
It's fine if it wraps. It's just not going to take as much space.

src/release/eol.rs
```diff
@@ -0,0 +1,51 @@
...
+            EolStatus::Imminent
+        } else {
+            EolStatus::Ok
+        }
+    }
+}
+
+#[inline]
+fn impending(current: Date<Utc>, eol_year: i32, eol_month: u32) -> bool {
+    current.year() == eol_year && current.month() == eol_month
```
leviport: I think this check isn't going the full month in advance since it's checking only within `eol_month`. For example, 19.10 EOL's on July 17, 2020. When I set the system date to June 30, 2020, I do not receive the EOL message, but going to July 1st makes it show up.

mmstick:
Adding a custom label to the button with a margin fixes the button size issue.

maria-komarova:
@mmstick can I take a look?

mmstick:
With a label that has the `margin` property set to `4`:

Screenshot from 2020-01-09 14-05-52

maria-komarova:
Looks good to me! And let's leave the notification as is now. Shrinking the wording there might make the meaning cryptic and users will either click on it right away or go check it in the list of notifications where they will most likely hover over it if they are curious. And clicking would take them to the upgrade page with more explanations anyway.

src/release/eol.rs
mmstick: I'll convert the EOL date to a `Date<Utc>`, and then check if the duration between the EOL and current date is more than 30 days.

maria-komarova:
A version of the notification for after EOL if Upgrade is still available:
notification-oldOS-EOL (1)
If Upgrade is not available, then say "Install Pop!_OS 19.10" instead of "Upgrade to" in the last sentence. And if this happens, we should also include the same text in the OS Upgrade field in the Settings. I don't think we can just not show OS Upgrade if we showed it before so we will have to provide this message about Support ended etc.

src/release/eol.rs
mmstick: This should be fixed now

mmstick:
Post-EOL notification is now included

src/cli/mod.rs
```diff
@@ -518,6 +529,42 @@ fn installed_after_release(next: &str) -> bool {
...
+        Ok(eol) => {
+            let (y, m, d) = eol.ymd;
+            match eol.status() {
+                EolStatus::Exceeded => {
+                    return (
+                        fomat!("Support for Pop!_OS " (current) " has ended"),
+                        fomat!(
+                            "Security and application updates are no longer provided "
+                            "for Pop!_OS " (current) ". Upgrade to Pop!_OS " (next)
+                            " to keep your computer secure."
```
brs17: I am not seeing this code path executed when the date is beyond that of the EOL date.

gtk/src/gtk_utils.rs
```diff
@@ -0,0 +1,7 @@
+use gtk::prelude::*;
+
+pub fn scale_label(label: &gtk::Label, scale: f64) {
+    let list = label.get_attributes().unwrap_or_else(|| pango::AttrList::new());
+    list.insert(pango::Attribute::new_scale(0.9).expect("new scale returned null"));
```
crawfxrd: `new_scale(0.9)` → should use `scale`

mmstick: Fixed

src/cli/mod.rs
mmstick: This is perplexing because the code that returns this is:

```rust
let (year, month, day) = self.ymd;
let eol = ymd_to_utc(year as i32, month, day);
let current = Utc::now().date();

if current >= eol {
    EolStatus::Exceeded
} else if imminent(current, eol) {
    EolStatus::Imminent
} else {
    EolStatus::Ok
}
```

If the current date is greater than, or equal to, the eol date; it should be returning `Exceeded`.

src/cli/mod.rs
```diff
@@ -518,6 +529,42 @@ fn installed_after_release(next: &str) -> bool {
     false
 }
 
+fn notification_message(current: &str, next: &str) -> (String, String) {
+    match EolDate::fetch() {
+        Ok(eol) => {
+            let (y, m, d) = eol.ymd;
```
crawfxrd: nit: scope can be reduced to `EolStatus::Imminent` branch.

src/cli/mod.rs
crawfxrd: Would it be feasible to add tests for that function?

mmstick: Tests can be added to the eol module to test the fetch method, yes.

PR #80, opened 2020-01-16
Refresh OS Support

mmstick:


leviport:
The refresh button I'm seeing is green:
refresh-button-green

and clicking on it goes to a dialog with a red button:
refresh-button-clicked

The green button -> red button sequence caught my attention, so I wanted to bring it up. To me it seems like the refresh button should be clear colored. @maria-komarova what are your thoughts on this?

maria-komarova:
@leviport There was a clear button on the original designs. You are right, it shouldn't be green. Thanks for bringing this to my attention. @mmstick can we change it?

brs17:
Refresh button is still green at all times.

I also did this install with the latest 18.04 iso (the one from the website) and I still had the button saying I could upgrade the recovery partition. What was it updating?
Screenshot from 2020-01-29 11-45-20

mmstick:
I haven't changed the button styles.

The button for upgrading the recovery partition updates the recovery partition with the latest version of the recovery partition, if it is not already updated to the defined "urgent" build ID in the source code.

brs17:
> The button for upgrading the recovery partition updates the recovery partition with the latest version of the recovery partition, if it is not already updated to the defined "urgent" build ID in the source code.

I thought the "urgent" ID was based on released .isos which we don't yet have implemented (so there should be no available updates at all).

mmstick:
It's currently hard-coded for testing, and we may or may not use the release API at all for this.

brs17:
I just did an install with a recovery partition without the refresh os feature and I believe I should have seen this button logic:
LTS-notifications

 when I just saw two green buttons under OS Recovery (for recovery partition update and for refresh os)

mmstick:
So it's allowing to boot into the recovery partition, even without updating the recovery partition?

brs17:
> So it's allowing to boot into the recovery partition, even without updating the recovery partition?

Yes sir.

leviport:
I'm not quite done testing yet, but there is one thing I noticed that might need a little attention. Right now updating the recovery partition shows the download progress, but once the download is complete, the message doesn't change until the recovery update is completed. It takes about another 20-30 seconds to make the update, so showing a completed download for that time makes it feel like it hung.

leviport:
It also seems that if the Gnome Control Center window is closed while a recovery update is downloading, it keeps downloading in the background. If it finishes the update while G-C-C is closed, I see a message that says "Failed to check for recovery updates", but the Refresh button is active and can be clicked. Continuing to update the recovery partition in the background might not be a bad thing, but it should at least report the status correctly once G-C-C is reopened.

leviport:
Couple more things: 

- If I click Refresh and reboot to the recovery partition, then decide I don't want to refresh, I cannot boot back into my installed OS without selecting it in my bootloader menu. Maybe the action that resets this runs at boot, and since my recovery partition doesn't have the patch until I add it manually, it never gets reset?

- After clicking Refresh and rebooting to the recovery partition, the installer does take me to the refresh OS part as expected. There is also a "Cancel Installation" button that I can click, which takes be back to the screen with "try demo mode", "clean install", "refresh install", and "custom (advanced)". This screen has a "Back" button which is clickable but doesn't do anything. I'm not sure it should do anything, so maybe it should just be hidden when `MODE=refresh` in `recovery.conf`?

leviport:
* When updating the recovery partition, the "verifying" and "syncing" statuses are now reported which looks much better. 

* I can still close Gnome Control Center during the recovery update download and reopen it to find an update button shown again. Clicking the update button shows the download still in progress in the background. If I let the download complete before reopening, it shows me a message telling me that the recovery partition is up to date, which is better than the error it was showing previously. 
     - I'm not certain what the expected behavior is when G-C-C is closed mid-download, but I think the current behavior is fine. The recovery partition updates correctly even if a user can make something harmless but slightly weird happen by antagonizing G-C-C.

* After clicking Refresh, booting into the recovery partition, then rebooting without refreshing, I'm still seeing some inconsistent behavior. I'll continue looking into it, but right now it looks like systemd-boot is still "sticking" on the recovery partition if I don't go through the refresh process. The installer now doesn't seem to be re-entering refresh mode after such a reboot, so something might be getting reset.

mmstick:
I don't think distinst is setting it back to the default boot option

mmstick:
Static files for fetching the latest release are in. When  checking for a release with `pop-upgrade release check`, you'll get an output like so whenever there is an urgent release available:

```
      Current Release: 20.04
         Next Release: 20.10
New Release Available: false
  Urgent Recovery ISO: 2020-01-24 17:53:13
```

mmstick:
With the latest commit, connections to `api.pop-os.org` no longer occur. Both build and urgent information is now contained within RON files maintained in this repository. When releasing a new ISO, the `build` field should be updated with the URL, SHA, size of the ISO, and the build number.

## 18.04 (Intel)

```ron
#![enable(implicit_some)]
(
	build: (
		url: "https://pop-iso.sfo2.cdn.digitaloceanspaces.com/18.04/amd64/intel/59/pop-os_18.04_amd64_intel_59.iso",
		sha: "0ae2c20327bc1059892c9efea71b21753782979431091fa3da60e4eaa036db1c",
		size: 2256076800,
		build: 59
	),
	urgent: (
		date: "Fri, 31 Jan 2020 20:46:23 +0000",
		build: 59,
	)
)
```

RON requires `#![enable(implicit_some)]` at the top of the file when specifying that optional fields may be implicit. Which means not having to explicitly write `Some(( key: value, key: value ))` to denote that a field has a value.

## 20.04 (Intel)

When no release information is available, we can have the file written as `()`.

```ron
()
```

leviport:
- Once booted into the recovery partition, I see the "Refresh" part of the installer and a "Cancel Installation" button. Clicking the cancel button takes me to the part of the installer with "clean install" or "custom install". This screen has a "back" button, which now does nothing. From a design perspective, it makes me wonder if the "cancel installation" button should remove the refresh flags and reboot into the normal install. If not, we probably want to make sure the "back" button isn't there at all.

- I'm also still getting stuck booting to the recovery partition if I decide to reboot without going all the way through the refresh process. To get out of this, I either need to visit the systemd-boot menu or completely refresh my OS. I think making it easier to back out at this stage is a good thing to do, even though there have been opportunities to do so before this point in the process.

maria-komarova:
Speaking about the first point, I would agree that "cancel installation" button should reboot into the normal install. Thinking about the user path, I wouldn't expect "cancel" button to take one to more options that do not feel like canceling. It results in a very meandering user path. If we would like to give users an option to get to the clean install from the Settings, maybe we should consider adding it to the page or some other alternative.

mmstick:
Yes, any changes here won't affect the installer's behavior. I will be looking into the installer and distinst today. Any issues with the GTK widget and upgrade daemon that I've missed?

mmstick:
Pop ISOs are now cached in `/var/cache/pop-upgrade/`, alongside the release cache. It is thereby possible to update the recovery partition across many systems on the same network without needing to download it on each system — provided that each system has the cache directory mounted to the same file share. Each ISO is named as `${version}-${kind}.iso`, so it is possible to cache ISOs for different versions and kinds.

All files on the recovery partition, besides `recovery.conf`, are now deleted before extracting a new ISO to the recovery partition. Files are then copied over in Rust, instead of relying rsync. The casper directory is recreated.

brs17:
> After selecting "Refresh" and then the "Reboot and Refresh" button, the system reboots but returns to the login screen. Seems to not be setting the recovery partition as the next boot default.

Ahh, it appears the default entry had been set in the systemd-boot menu. Since that is stored differently, that would explain why it wasn't rebooting into the recovery.

brs17:
I just did an upgrade from 19.10->20.04 and it said the Recovery partition was up to date, when I tried a Refresh OS install, it took me to the recovery partition which was still 19.10.

jackpot51:
Until we whitelist 20.04 I believe that is expected

brs17:
> Until we whitelist 20.04 I believe that is expected

Since we might expect a user is running a beta version of Pop!_OS or something similar in the future, shouldn't we more elegantly handle running a version without the appropriate recovery partition in place? Perhaps default to a similar state as when there is no recovery partition?

WatchMkr:
I don't think it's required for release because 20.04 will be out. We can address it when we open up the gorilla.

PR #83, opened 2020-01-24
fix(daemon): Require supported releases to be hardcoded

mmstick:


mmstick:
Latest change refactors all release check logic into a new `src/release/check.rs` module, which hardcodes the supported releases. The builds API is now only used to check for a connection to the builds server, and to fetch the build ID.

src/release/check.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+pub fn next(development: bool) -> Result<ReleaseStatus, VersionError> {
+    let current = Version::detect()?;
+
+    let next: &str;
+
+    let status = match (current.major, current.minor) {
+        (18, 4) => {
+            next = if development { "20.04" } else { "19.04" };
```
jackpot51: else { "19.10" }

src/release/check.rs
```diff
@@ -0,0 +1,117 @@
...
+            ReleaseStatus {
+                build: Release::build_exists(next, "intel").into(),
+                current: "18.04",
+                is_lts: true,
+                next,
+            }
+        }
+
+        (19, 10) => {
+            next = "20.04";
```
jackpot51: if development {  "20.04" } else { "19.10" or appriopriate error }

src/release/check.rs
```diff
@@ -0,0 +1,117 @@
...
+                is_lts: false,
+                next,
+            }
+        }
+
+        (20, 4) => ReleaseStatus {
+            build:   BuildStatus::Blacklisted,
+            current: "20.04",
+            is_lts:  true,
+            next:    "20.10",
```
jackpot51: Wouldn't this allow 20.04 to upgrade to 20.10 immediately?

mmstick: Not unless `build` contains a `BuildStatus::Build` signifying that an ISO is available.

jackpot51: That is going to happen eventually so shouldn't it also do what is done in the 19.10 listing?

PR #84, opened 2020-01-27
fix(packaging): Debian package did not contain systemd service and timer mer for notification checks

mmstick:


mmstick:
This PR doesn't make any changes to the client or the daemon (just adds the systemd notification service script + timer), so an error caused by this shouldn't be possible.

mmstick:
I believe I've found the problem

mmstick:
The systemd service for `pop-upgrade` was not being included in the package

leviport:
I'm still seeing an error after the latest commits. It looks like it's related to the systemd service.
upgrade-error

I also wasn't able to see notifications after spoofing the install date by changing the clock and touching `/etc/machine-id`. I'm guessing that's also related.

mmstick:
That one should be fixed now, since the files are moved to `/usr/lib`.

leviport:
It's still having an issue with that file. Looking in that directory though, I'm finding the file `/usr/lib/systemd/system/pop-upgrade-init.service` file, but not the one at `/usr/lib/systemd/system/system-update.target.wants/pop-upgrade-init.service` like the notification mentions.
Screenshot from 2020-01-30 14-00-19

mmstick:
That one's also being placed in the wrong location, as a symbolic link.

PR #102, opened 2020-04-30
Whitelist 20.04 upgrades, and blacklist 20.10

mmstick:


leviport:
I saw this, but changelogs get added in a different spot, right?
nochangelog

Also I won't be able to test that 20.10 is blacklisted until it builds for focal

mmstick:
Yes, this enables the upgrades, but doesn't add any changelogs

leviport:
On 18.04 I didn't get a notification, and it looks like notifications are dismissed by default?
dismissed

PR #112, opened 2020-05-28
Dev Branch

mmstick:


leviport:
I'm getting an upgrade failure when trying to go from 18.04 to 20.04 on serw12. It did safely recover afterwards, leaving sources.list in a good state. 

upgrade-encountered-issue-06-11-20

It looks like the failure is at line 14056 in the [journalctl log](https://github.com/pop-os/upgrade/files/4766862/journalctl.txt). Something about a checksum mismatch

I like the additional detail in the error message though! We might want to think about a better way to show that info than that big red block, but it works for now. We should get UX's opinions on it at some point though.

jacobgkau:
I tested this with a fresh install from 18.04 to 20.04 on a gaze14. I did not see any errors. The system did hang at "Upgrade complete. Now rebooting" for over five minutes. When I manually shut the machine off and turned it back on, it booted into 20.04, and things appear to be in a healthy state.

mmstick:
@jacobgkau I see what you mean. I've re-added the statement for rebooting on successful upgrades.

src/repair/sources.rs
```diff
@@ -114,19 +121,33 @@ fn insert_entry<P: AsRef<Path>>(
...
-         deb http://apt.pop-os.org/proprietary {0} main\n",
-         release
-    ))?;
+pub fn create_new_sources_list(release: &str) -> io::Result<()> {
+    fs::write(
+        MAIN_SOURCES,
+        format!(
+            r#"# Ubuntu Repositories
+
+deb http://us.archive.ubuntu.com/ {0} restricted multiverse universe main
```
jacobgkau, line 130: I'm now getting `unable to upgrade to next release: failed to update sources` on every attempt, even though `sudo apt update` is not showing any errors. It looks like the four `us.archive.ubuntu.com` links need to have /ubuntu at the end like they used to.

jacobgkau: Fixed with https://github.com/pop-os/upgrade/pull/112/commits/2a8ffe53ef4e44b9ac25ed088eb9f223738b0392.

PR #122, opened 2020-09-14
Update ubuntu-version to support 20.10 groovy

jackpot51:
This will allow upgrades to 20.10 using the --force-next flag


leviport (approved):
Pop-upgrade seemed to work properly with this patch, despite Groovy not being all there yet. An 18.04 to 20.04 upgrade also worked properly with this patch.

mmstick (approved):

Issue #116, opened 2020-07-07
Invalid third-party repo/PPA stops upgrade from proceeding


jacobgkau:
If a user has added a repository to their system that no longer exists (or never existed in the first place), pop-upgrade will stop the upgrade. While the new error messages give the user more of a hint about what the issue is, they still don't mention the specific repository that's causing the issue, nor is there an easy way to disable the repository (such as a button or instructions to open Repoman.)

It seems like if the upgrader disables PPAs by default, a bad PPA should not stop the upgrader from proceeding, it should just be disabled (or provide guidance about what/how to disable.)

Steps to reproduce:

* Perform clean install of Pop!_OS 18.04 LTS and install all updates
* Run the following commands:
```
git clone https://github.com/pop-os/pop
cd pop/scripts
./apt add bad-ppa
```
* Open Settings and attempt to upgrade to 20.04 LTS

Error message with no mention of "bad-ppa":

![upgrade-i116-1](https://user-images.githubusercontent.com/7199422/86812998-e3c30100-c03c-11ea-9ca6-5a5786ac2587.png)

Error message in `sudo apt update`, which is very obvious to me but many users do not know how to read/interpret:

![upgrade-i116-2](https://user-images.githubusercontent.com/7199422/86813051-f1788680-c03c-11ea-9b6a-0b5ca79c6307.png)

mmstick:
We'll need to scrape the output of `apt update` to look for these 404 errors. It shouldn't be that difficult.

jacobgkau:
We now disable all third-party sources at the beginning of the upgrade process, so this is no longer an issue.

PR #124, opened 2020-10-15
fix: Source list backups now comment third party sources

mmstick:
Changes the behavior from copying sources to a `.bak` directory, to creating `.save` copies and overwriting the original with `#` before deb lines.


jacobgkau (approved):
Focal installs from the iso with the Pop PPA in `sources.list.d/system76-ubuntu-pop-focal.list`. After running an upgrade to Groovy, pop-upgrade places the Pop PPA in the main `sources.list` and (after this change) also leaves the same PPA (disabled) in `sources.list.d/system76-ubuntu-pop-focal.list`. This means if I open Repoman after the upgrade, I see the Focal version of the Pop PPA showing up as "disabled" in the Extra Sources tab, with the enabled Groovy version being in the "Legacy System Sources" entry.

https://github.com/pop-os/upgrade/pull/125 is going to add the system sources in deb822 format instead of one-liner format, and will also place the Pop PPA inside of `sources.list.d/pop-os-ppa.list` instead of `sources.list` as of right now. That's going to end up showing the Pop PPA twice in Repoman, the enabled Groovy version and the disabled Focal version; for cleanliness, we may want to consider removing `sources.list.d/system76-ubuntu-pop-*.list` upon upgrade, so only the most recent Pop PPA is listed. (If the original install was an older version than Focal, then the Focal PPA might be active inside of a differently-named file, such as my work machine, which is currently using `sources.list.d/system76-ubuntu-pop-disco.list` with a focal line inside.)

One remaining "gotcha" with the new flow is that the sources in Repoman will be pre-populated with the incorrect (outdated) release name, so instead of just flipping the switch back to "Enabled", users will need to know the correct name for their current release. Not sure if we'd want to try and update the release name at the same time we're disabling the line to make it easier to re-enable afterwards-- there could still be cases where the PPA hasn't added the new release yet, some repos don't use the standard release names, etc.

I'm approving this since it is working as intended and nothing I've described above is broken, just a little messy/confusing. This does not address https://github.com/pop-os/upgrade/issues/116; a bad PPA/repo being configured will stop the upgrade from proceeding (before it gets to the point where it would disable the bad third-party PPA/repo.)

mmstick:
#116 is more involved because there's no clean way to handle it besides telling the user about the problem. I'm working on a solution for that today. Once #125 is complete we can do what's needed to organize the mess.

src/release/repos.rs
```diff
@@ -1,29 +1,100 @@
...
 
-    Ok(())
+/// For each `.list` in `sources.list.d`, add `#` to the `deb` lines.
+pub fn disable_third_parties() -> anyhow::Result<()> {
+    let dir = fs::read_dir(PPA_DIR).context("cannot read PPA directory")?;
+    iter_files(dir, |entry| {
+        if entry.file_name().to_bytes().ends_with(b".list") {
+            let path = entry.path();
+
+            if let Some(fname) = path.file_name() {
```
jacobgkau, line 69: I had just looked at my `sources.list.d/system76-ubuntu-pop-focal.list` file, so I know it was present. After attempting to run an upgrade, I got this output:

```
system76@pop-os:~$ pop-upgrade release upgrade -f
Current Release: 20.04
Upgrading to: 20.10
New version available: false
Release upgrade status: release upgrade aborted: unable to disable third party repositories: failed to read /etc/apt/sources.list.d/system76-ubuntu-pop-focal.list
pop-upgrade: dbus client error: calling ReleaseUpgradeFinalize method failed
```

The `sources.list.d/system76-ubuntu-pop-focal.list` file is gone, and trying to upgrade again gives the same output.

- We probably shouldn't fail if it's not found. We've been inconsistent with where we're placing the PPA, and there could be valid configurations where this file isn't present to remove.
- This section to remove the system76-ubuntu-pop file should probably be below the section where sources are disabled (I would assume it's easier to disable everything in sources.list.d first and then remove this particular file, rather than trying to remove this one first and then disable everything except for this one, but whichever way works should be fine.)

mmstick: Oh I see. It's removed and then attempted to back up.

jacobgkau:
I'm currently getting this:

```
system76@pop-os:~$ pop-upgrade release upgrade -f
Current Release: 20.04
Upgrading to: 20.10
New version available: false
Event: removing deprecated and/or conflicting packages
Event: updating package lists
Release upgrade status: release upgrade aborted: failed to update package lists for the current release: command failed with exit status: exit code: 100
pop-upgrade: dbus client error: calling ReleaseUpgradeFinalize method failed
```

I currently have an invalid repo configured in the main `sources.list` (the same one I used as an example in #116). I also added a valid PPA in `sources.list.d` to check if it would be disabled, and it doesn't look like it is.

mmstick:
The system sources list isn't touched at all until an upgrade is ready, since we need to update the current OS release before we upgrade to the next. So it would take a lot more effort to find bad PPAs (if they're even bad and not just temporarily unavailable).

mmstick:
Just going to disable any `deb` line that lacks Ubuntu, System76, or Pop from their names

mmstick:
> I was imagining we just run create-new-sources-list twice throughout the process,

Not sure why I didn't think of that. Seems I need to make some coffee today.

> Here's what sudo apt update shows:

The logs from pop-upgrade's daemon would be more useful. You can see these streaming when running this in another terminal:

```sh
journalctl -o cat -f -u pop-upgrade
```

But it looks like apt-get update is being called somewhere before the source lists are modified.

jacobgkau:
> The logs from pop-upgrade's daemon would be more useful. You can see these streaming when running this in another terminal:
> 
> ```shell
> journalctl -o cat -f -u pop-upgrade
> ```
> 
> But it looks like apt-get update is being called somewhere before the source lists are modified.

Got it, here's the output of that: [upgrade-pr124-output4.txt](https://github.com/pop-os/upgrade/files/5394654/upgrade-pr124-output4.txt)

I do see the line `mod.rs:341: disabling third party sources` before it updates the package lists.

jacobgkau:
With `9f40aca`, I'm still getting `failed to update package lists for the current release: command failed with exit status: exit code: 100`.

If I put a bad source into the main `sources.list` file, I'm seeing it get removed after attempting to run pop-upgrade, so that part's working correctly now. The bad PPA inside of `sources.list.d` is also being commented out correctly. It appears the issue is that `release` is being set to `20.04` instead of `focal` in the regenerated sources.list file:

```
system76@pop-os:~$ cat /etc/apt/sources.list
# Ubuntu Repositories

deb http://us.archive.ubuntu.com/ubuntu/ 20.04 restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04 restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-updates restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-updates restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-security restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-security restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-backports restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-backports restricted multiverse universe main

# Pop!_OS Repositories

deb http://ppa.launchpad.net/system76/pop/ubuntu 20.04 main
deb-src http://ppa.launchpad.net/system76/pop/ubuntu 20.04 main

deb http://apt.pop-os.org/proprietary 20.04 main
```

When creating the new sources.list file, it looks like we convert from the number to the name shortly before: https://github.com/pop-os/upgrade/blob/9f40acacc18c0f49869da558c000eafa7ca56db3/src/release/mod.rs#L250 

So we probably need to do that here as well, or we need to pass the codename into the `backup`, `disable_third_parties`, and `repair` functions instead of the version number.

jacobgkau (changes requested):
With `9f40aca`, I'm still getting `failed to update package lists for the current release: command failed with exit status: exit code: 100`.

If I put a bad source into the main `sources.list` file, I'm seeing it get removed after attempting to run pop-upgrade, so that part's working correctly now. The bad PPA inside of `sources.list.d` is also being commented out correctly. It appears the issue is that `release` is being set to `20.04` instead of `focal` in the regenerated sources.list file:

```
system76@pop-os:~$ cat /etc/apt/sources.list
# Ubuntu Repositories

deb http://us.archive.ubuntu.com/ubuntu/ 20.04 restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04 restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-updates restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-updates restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-security restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-security restricted multiverse universe main

deb http://us.archive.ubuntu.com/ubuntu/ 20.04-backports restricted multiverse universe main
deb-src http://us.archive.ubuntu.com/ubuntu/ 20.04-backports restricted multiverse universe main

# Pop!_OS Repositories

deb http://ppa.launchpad.net/system76/pop/ubuntu 20.04 main
deb-src http://ppa.launchpad.net/system76/pop/ubuntu 20.04 main

deb http://apt.pop-os.org/proprietary 20.04 main
```

When creating the new sources.list file, it looks like we convert from the number to the name shortly before: https://github.com/pop-os/upgrade/blob/9f40acacc18c0f49869da558c000eafa7ca56db3/src/release/mod.rs#L250 

So we probably need to do that here as well, or we need to pass the codename into the `backup`, `disable_third_parties`, and `repair` functions instead of the version number.

mmstick:
At some point I should use newtypes to protect against using a version string in a function that requires the codename string.

jacobgkau:
Looks good now. Successfully moved my old `sources.list` to `sources.list.d` and generated a new `sources.list`, and commented out my PPAs in `sources.list.d`, eliminating all invalid sources and allowing the upgrade to proceed while also allowing me to re-enable the `sources.list.d` PPAs in Repoman after the upgrade (except for the old Pop PPA file, which was correctly removed to prevent duplicates in Repoman.)

Also tested a second time after moving the Pop PPA into the main `sources.list` file and removing its `sources.list.d` file manually before running `pop-upgrade`, everything still worked.

I did notice while watching the logs that it looks like we're checking for the proprietary repo and the Pop PPA before we recreate `sources.list`:

```
Oct 19 10:05:00 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/daemon/mod.rs:641: upgrading release from 20.04 to 20.10, with 1
Oct 19 10:05:00 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/daemon/mod.rs:237: attempting release upgrade, using a offline upgrade
Oct 19 10:05:00 pop-os pop-upgrade[5628]: killing io.elementary.appcenter
Oct 19 10:05:00 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/repair/mod.rs:31: performing release repair
Oct 19 10:05:00 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/repair/sources.rs:44: ensuring that the proprietary pop repo is added
Oct 19 10:05:00 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/repair/sources.rs:82: PPA system76/pop found: not adding
Oct 19 10:05:01 pop-os pop-upgrade[6300]: Reading package lists...
Oct 19 10:05:01 pop-os pop-upgrade[6300]: Building dependency tree...
Oct 19 10:05:01 pop-os pop-upgrade[6300]: Reading state information...
Oct 19 10:05:01 pop-os pop-upgrade[6300]: 0 upgraded, 0 newly installed, 0 to remove and 264 not upgraded.
Oct 19 10:05:02 pop-os pop-upgrade[6310]: pop-upgrade set on hold.
Oct 19 10:05:02 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/release/mod.rs:331: creating backup of source lists
Oct 19 10:05:02 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/release/repos.rs:51: removing old backup at /etc/apt/sources.list.save
Oct 19 10:05:02 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/release/repos.rs:54: creating backup of /etc/apt/sources.list to /etc/apt/sources.list.save
Oct 19 10:05:02 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/release/mod.rs:334: disabling third party sources
Oct 19 10:05:02 pop-os pop-upgrade[6321]: Reading package lists...
Oct 19 10:05:02 pop-os pop-upgrade[5628]: [INFO ] pop_upgrade src/daemon/mod.rs:457: release upgrade: removing deprecated and/or conflicting packages
Oct 19 10:05:03 pop-os pop-upgrade[6321]: Building dependency tree...
```

That seems unnecessary, but shouldn't cause any problems. However, what I'm more concerned about is `repair/sources.rs` also being responsible for switching from `us.archive.ubuntu.com` to `old-releases.ubuntu.com` for EOL releases (reading through it [here](https://github.com/pop-os/upgrade/blob/upgrade/src/repair/sources.rs#L44)). If we run the repair _before_ we regenerate `sources.list`, would we end up overwriting `old-releases` with `us.archive` again because `create_new_sources_list` is hardcoded to `us.archive`?

@mmstick Is this a valid concern?

mmstick:
That's true and valid. :thinking:
I'll add a check and make sure old-releases is replaced for an upgrade

jacobgkau:
Would it be easier to just remove the if statement [here](https://github.com/pop-os/upgrade/blob/upgrade/src/repair/sources.rs#L35) so the new sources.list file always gets created at the beginning of the `repair` we're already running? Just an idea.

mmstick:
`pop-upgrade release repair` calls this and would wipe out sources even if not upgrading.

mmstick:
Though I would like to separate system repairs to a new project eventually, instead of bundling it in pop-upgrade

mmstick:
`repair::sources` module will be no more now. The sources list will be switched to `old-releases` when the release's repository exists on Ubuntu's old-releases archive.

src/release/mod.rs
```diff
@@ -324,8 +312,7 @@ impl DaemonRuntime {
     ) -> RelResult<()> {
         self.terminate_background_applications();
 
-        // Check the system and perform any repairs necessary for success.
-        repair::repair().map_err(ReleaseError::Repair)?;
+        let from_codename = from.parse::<Codename>().unwrap();
```
jacobgkau, line 315: Failing on [this line](https://github.com/pop-os/upgrade/blob/da0d8153e879c73e89ec674f06b973bf933a4c55/src/release/mod.rs#L315):

```
Oct 19 15:08:36 pop-os pop-upgrade[7535]: killing io.elementary.appcenter
Oct 19 15:08:36 pop-os pop-upgrade[7535]: thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: NotFound', src/release/mod.rs:315:29
Oct 19 15:08:36 pop-os pop-upgrade[7535]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The backtrace doesn't look too useful, but here it is: [backtrace.txt](https://github.com/pop-os/upgrade/files/5404795/backtrace.txt)

jacobgkau (changes requested):

src/release/mod.rs
mmstick: Oops. I know what happened

jacobgkau (approved):
Both example configs are back to working smoother than stable.

jacobgkau:
Just noticed that if I run `pop-upgrade release repair` once I'm already on Groovy, the proprietary repo is added to `sources.list` and not removed from its deb822 location at `sources.list.d/pop-os-apps.sources`. It's removed again next time `pop-default-settings` is reconfigured. (Maybe https://github.com/pop-os/upgrade/pull/125 should deal with that, since it's bringing deb822 into pop-upgrade?)

mmstick:
Yeah this isn't yet DEB822 aware

mmstick:
I'll merge for now on the basis that we're going to get DEB822 improvements immediately after.

PR #125, opened 2020-10-15
fix(sources): upgrade to DEB822 sources

isantop:
On upgrades, replace the one-line format sources with DEB822-format
sources instead. This prevents pop-default-settings from needing to
do so.


src/repair/sources.rs
```diff
@@ -122,32 +125,77 @@ fn insert_entry<P: AsRef<Path>>(
 }
 
 pub fn create_new_sources_list(release: &str) -> io::Result<()> {
```
mmstick, line 127: You can change `-> io::Result<()>` to `-> anyhow::Result<()>` for those errors below.

src/repair/sources.rs
```diff
@@ -122,32 +125,77 @@ fn insert_entry<P: AsRef<Path>>(
...
-    fs::write(
-        MAIN_SOURCES,
-        format!(
-            r#"# Ubuntu Repositories
+    // Switchover is the release where DEB822-format sources were adopted. 
+    // Currently 'groovy'
+    let switchover = "groovy";
+    let new_timestamp = release.parse::<Codename>().release_timestamp()?;
+    let switchover_timestamp = switchover.parse::<Codename>().release_timestamp()?;
```
mmstick, line 132: I think you need to use `?` like so `<Codename>()?.release_timestamp();`

isantop:
After testing, this branch successfully upgrades 18.04 to 20.04 without converting to DEB822 sources, and converts sources to DEB822 when upgrading to groovy. I believe checking the timestamps of the releases is the best way to handle this checking.

mmstick (approved):

jacobgkau (approved):
Working as advertised; on a Focal -> Groovy upgrade, the re-generated Focal list is still in the old format, while the generated Groovy list is in deb822 format. For Bionic -> Focal, both versions use the old format.

For Focal -> Groovy, I am no longer seeing the behavior I described in [this comment](https://github.com/pop-os/upgrade/pull/124#issuecomment-712489671). `pop-upgrade release repair` doesn't seem to do anything with missing sources (if the proprietary repo is missing, it's added added back when `pop-default-settings` is reconfigured; if the PPA is missing, it can be added back manually with `sudo add-apt-repository ppa:system76/pop`.) It doesn't look like this PR did anything to affect this, so it's possible I was using the released version of pop-upgrade when I noticed that behavior since the staging repo would have been removed during the upgrade.

mmstick:
Yeah this PR doesn't add anything to `release repair`, since source repository manipulation got scrapped in the last PR. It could be re-added though if there's a benefit to having this feature, though I think we need system repair functionality in a separate project.

PR #126, opened 2020-10-21
fix(repair): Add plain option to cryptswaps in crypttab

### LINKED ISSUE #123: 20.10: Decryption screen displays a cryptswap warning after upgrade from 20.04
After upgrades from 20.04, I'm seeing a warning that states, `cryptsetup: WARNING: cryptswap: couldn't determine device type, assuming default (plain).` The drive still unlocks and the OS boots normally, but the warning may be unsettling for users.

Perhaps this isn't the correct repo for this issue, but I'm not seeing it on fresh Groovy installs, so I figured it could be related to the upgrade process.

PXL_20201009_222221225

mmstick:
Closes #123

### DIFF
--- a/src/release/mod.rs
+++ b/src/release/mod.rs
@@ -313,7 +313,8 @@ impl DaemonRuntime {
         self.terminate_background_applications();
 
         let from_version = from.parse::<Version>().expect("invalid version");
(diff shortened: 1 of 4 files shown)


leviport (approved):
Looks like that did the trick!

PXL_20201021_190839858

ids1024 (approved):

PR #127, opened 2020-10-22
improv: Update dependencies & Async / package-fetching updates

mmstick:
Contains significant improvements to major portions of the daemon, will majorly cut down dependencies, and replace outdated ones, and leave us with code that's easier to maintain. This contains a transition to using async functions and methods, which allows us to do some things concurrently without needing additional threads. The previous code that was being used to fetch packages has been completely deprecated. The outdated code used to fetch ISOs will be replaced. DBus support will be able to move forward with becoming async.


mmstick:
I just had a successful upgrade with the new code. I'm going to continue testing different scenarios to make sure everything works before tomorrow morning.

mmstick:
Okay, this is ready for testing now.

jacobgkau:
When I tried testing this, pop-upgrade reset my sources (losing the `upgrade-upgrades` staging PPA) and then updated itself back to the stable version, then it failed to finalize the upgrade. Updating itself _after_ resetting the sources does seem correct (although it will make future updates to pop-upgrade difficult to test.) I'm not sure what caused the failure, but I saw a prompt to reboot after pop-upgrade was finished updating itself. Here's the terminal output I saw: [upgrade-pr127-terminal.txt](https://github.com/pop-os/upgrade/files/5476917/upgrade-pr127-terminal.txt)

Here's the daemon logs: [upgrade-pr127-daemon.txt](https://github.com/pop-os/upgrade/files/5476918/upgrade-pr127-daemon.txt)

When I tried adding the `upgrade-upgrades` staging PPA back to try a second time, I saw that I'd been left with broken dependencies:

```
The following packages have unmet dependencies:
 libpop-upgrade-gtk : Depends: pop-upgrade (= 0.1.0~1604294004~20.04~d0e5ae8) but 0.1.0~1603472643~20.04~05600dc is installed
```

So:
- If we're going to reset sources before checking if `pop-upgrade` has an update available (which might be necessary if we want to continue avoiding bad PPAs stopping an upgrade), then we might need a flag to disable that feature for testing.
- It looks like we might need to check if `libpop-upgrade-gtk` has an update at the same time as `pop-upgrade` (or handle that in a dependency-sensitive way) to avoid leaving the package manager in a bad state.
- The upgrade process should ideally continue after the daemon has been restarted and should not display the reboot message that early.

mmstick:
Sounds like it didn't get upgraded before the upgrade was initiated.

jacobgkau:
When I upgrade now, I see the staging repo get removed (leaving a different candidate version than the one I'm running, `d43e009`), but the candidate version isn't installed during the process, presumably because the candidate didn't change until after the `pop-upgrade` update check was complete. The upgrader is still unfazed by a bad repo being present.

Once I've rebooted into the system after the upgrade, I notice that when I run `pop-upgrade release upgrade`, I get the `Checking if pop-upgrade requires an update` message, but it's not actually updated:

```
system76@pop-os:~$ apt policy pop-upgrade
pop-upgrade:
  Installed: 0.1.0~1604341397~20.04~d43e090
  Candidate: 0.1.0~1603472643~20.10~05600dc
  Version table:
 *** 0.1.0~1604341397~20.04~d43e090 100
        100 /var/lib/dpkg/status
     0.1.0~1603472643~20.10~05600dc 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main amd64 Packages
system76@pop-os:~$ pop-upgrade release upgrade
Current Release: 20.10
Upgrading to: 21.04
New version available: false
no release available to upgrade to
system76@pop-os:~$ apt policy pop-upgrade
pop-upgrade:
  Installed: 0.1.0~1604341397~20.04~d43e090
  Candidate: 0.1.0~1603472643~20.10~05600dc
  Version table:
 *** 0.1.0~1604341397~20.04~d43e090 100
        100 /var/lib/dpkg/status
     0.1.0~1603472643~20.10~05600dc 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main amd64 Packages
```

@mmstick What conditions need to be met for `pop-upgrade` to actually get updated if an update is available? (I see the package lists being updated when I run `pop-upgrade release upgrade` in the journalctl logs, immediately followed by the release check.) If someone has a version new enough to self-update but too old to install a recent release, is the update going to be installed before the release check (which would cut down on the occasional issue/question where someone forgets to install updates before trying to upgrade to a brand new release), or will `pop-upgrade` only update itself if it knows there's a new Pop!_OS release available?

mmstick:
It installs `pop-upgrade` (which automatically restarts the daemon on upgrade) only if `apt list --upgradable` returns a `pop-upgrade` package line.

mmstick:
Should I change it to compare the `apt-cache policy` Installed / Candidate lines?

jacobgkau:
> It installs `pop-upgrade` (which automatically restarts the daemon on upgrade) only if `apt list --upgradable` returns a `pop-upgrade` package line.

I see, mine was showing up as an available downgrade rather than an upgrade, so it didn't show up in that command's output.

> Should I change it to compare the `apt-cache policy` Installed / Candidate lines?

As long as the check is performed before resetting the source lists (so staging repos still work), that might be a good idea to avoid corner cases like the one above, unless there are any drawbacks that I'm not thinking of. Other tools such as the pop-shop will install available downgrades when performing their update functions, so it's not like we would rely on the current behavior. I'm struggling to think of a scenario where it would cause an issue, and it would make this easier to test.

mmstick:
Fixing it to use `apt-cache policy`, since that's more reliable

jacobgkau:
I just completed an upgrade with `aad9fe9`. At the beginning of the process, the update check didn't seem to cause any problems, as I was still running `aad9fe9` when the upgrade started. After the reboot, I ended up with the `pop-upgrade` package updated to the current Groovy release (`05600dc`), but `libpop-upgrade-gtk` was still on `aad9fe9`; additionally, `pop-upgrade` was installed but not configured. This meant I had to manually run `sudo dpkg --configure -a` and then `sudo apt install -f` to recover.

Terminal output (after upgrade): [upgrade-pr127-output.txt](https://github.com/pop-os/upgrade/files/5483043/upgrade-pr127-output.txt)
journalctl log: [upgrade-pr127-journalctl.txt](https://github.com/pop-os/upgrade/files/5483056/upgrade-pr127-journalctl.txt)
upgrade.log: [upgrade-pr127-upgrade-log.txt](https://github.com/pop-os/upgrade/files/5483046/upgrade-pr127-upgrade-log.txt)

For this upgrade, the only non-default package I installed before running the upgrade was `wine`. I've never seen an upgrade finish and not have a `pop-upgrade` update available immediately after, so I'm assuming the issue is that `pop-upgrade` somehow updated itself to its Groovy version and not that `libpop-upgrade-gtk` wasn't upgraded.

mmstick:
@jacobgkau I got it working. Post-upgrade, both packages are available as updates. No more `dpkg` conflicts.

gtk/src/events/mod.rs
```diff
@@ -395,7 +391,9 @@ fn release_upgrade_dialog(state: &mut State, widgets: &EventWidgets) {
     let dialog = UpgradeDialog::new(&state.upgrading_from, &state.upgrading_to);
 
     let answer = dialog.run();
-    dialog.destroy();
+    unsafe {
+        dialog.destroy();
```
ids1024, line 395: The gtk-rs release notes mention that the `.close()` method can be used instead of `.destroy()` for closing windows, which avoids `unsafe`.

https://gtk-rs.org/blog/2020/07/04/new-release.html

ids1024 (approved):

jacobgkau:
> Post-upgrade, both packages are available as updates.

That's not quite what I'm seeing. After starting on `3f2d44f`, after the upgrade is complete, I already have `20.10~05600dc` installed, with no updates available. Both `pop-upgrade` and `libpop-upgrade-gtk` are now on that version, and there are no dpkg issues, so this is fine as long as it's expected.

One other thing I ran into was this scenario:

- Add `upgrade-upgrades` staging repo
- Upgrade `pop-upgrade` and start a release upgrade
- `pop-upgrade` sees that it's up-to-date, so it sets itself on hold, resets the package lists, and starts upgrading the current release
- While updating packages for the current release, stop the `pop-upgrade` service (I initially did this because a download [hung](https://github.com/pop-os/upgrade/issues/131))
- Next time the release upgrade is attempted, `pop-upgrade` sees that it's "out of date" since the package lists were reset, but it's unable to update itself since it's still on hold (journalctl shows it's continuously trying again)

One way this could manifest itself in the real world (where someone hasn't added a staging repo) is if the upgrade is stopped during upgrading the current release, and then an update becomes available for pop-upgrade before the user tries again. It's unlikely, and since the GUI doesn't provide a way to stop an upgrade in progress anyway, the upgrade theoretically shouldn't stop once it's started, but it might be worth either unholding `pop-upgrade` before checking if it needs an update, or not checking for a self-update if it's already held.

Aside from that edge case, everything is looking good. I'm seeing successful upgrades now.

mmstick:
I'll fix the hold

jacobgkau:
The hold issue is still occurring on `5ed2a06` if pop-upgrade is stopped while fetching updated packages for the current release. This is the repeating output in `journalctl`:

```
Nov 09 11:26:23 pop-os pop-upgrade[22348]: Reading package lists...
Nov 09 11:26:24 pop-os pop-upgrade[22348]: Building dependency tree...
Nov 09 11:26:24 pop-os pop-upgrade[22348]: Reading state information...
Nov 09 11:26:24 pop-os pop-upgrade[22348]: 0 upgraded, 0 newly installed, 0 to remove and 252 not upgraded.
Nov 09 11:26:25 pop-os pop-upgrade[22355]: Reading package lists...
Nov 09 11:26:25 pop-os pop-upgrade[22355]: Building dependency tree...
Nov 09 11:26:25 pop-os pop-upgrade[22355]: Reading state information...
Nov 09 11:26:25 pop-os pop-upgrade[22355]: The following held packages will be changed:
Nov 09 11:26:25 pop-os pop-upgrade[22355]:   pop-upgrade
Nov 09 11:26:25 pop-os pop-upgrade[22355]: The following packages will be DOWNGRADED:
Nov 09 11:26:25 pop-os pop-upgrade[22355]:   libpop-upgrade-gtk pop-upgrade
Nov 09 11:26:25 pop-os pop-upgrade[22355]: 0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 250 not upgraded.
Nov 09 11:26:25 pop-os pop-upgrade[22355]: E: Held packages were changed and -y was used without --allow-change-held-packages.
```

mmstick:
I wonder what the held packages were. `apt-mark showhold` would say which ones. I'm guessing this happened because the PPA was removed and it's needing to downgrade, but there's an upgrade call that I made that doesn't permit downgrades.

jacobgkau:
> I wonder what the held packages were. `apt-mark showhold` would say which ones. I'm guessing this happened because the PPA was removed and it's needing to downgrade, but there's an upgrade call that I made that doesn't permit downgrades.

`apt-mark showhold` shows that the only held package is `pop-upgrade`. This is still happening with `49b100b`, with the `journalctl` output the same as my previous comment. It's not getting un-held before it tries to self-update.

mmstick:
Do you have the full log? Wondering why there's an upgrade happening when it's unheld on a restore

jacobgkau:
> Do you have the full log? Wondering why there's an upgrade happening when it's unheld on a restore

The last commit fixed the hold issue, it downgrades itself as expected now. I'm assuming the reason it wasn't working before was because pop-upgrade wasn't exiting cleanly, so it wasn't being "restored" (if it was being restored, I would also have expected the staging repo to be added back?)

The older version of `pop-upgrade` doesn't properly continue the upgrade after the daemon is restarted, so for the rest of the testing, I'm checking out and building this branch locally, then letting it self-update to the staging package (same version but different version number.)

Screenshot from 2020-11-10 09-14-16

The upgrade failed because `libpop-upgrade-gtk` was not self-upgraded at the same time as `pop-upgrade`, which created a dependency issue.

Screenshot from 2020-11-10 09-14-34

Full log: [upgrade-pr127-output.txt](https://github.com/pop-os/upgrade/files/5518622/upgrade-pr127-output.txt) (It looks like `libpop-upgrade-gtk` was at least downloaded, but `pop-upgrade` being restarted interrupted the process before it was installed.)

I do have `libpop-upgrade-gtk-dev` installed and I'm not sure if this happens without that installed yet, but it's in our regular PPA, so we should handle it.

jacobgkau:
I tried this again without `libpop-upgrade-gtk-dev` installed, and it still failed. `pop-upgrade` was upgraded, `libpop-upgrade-gtk` was not upgraded, and I was left with a dependency issue. [upgrade-pr127-nov11.txt](https://github.com/pop-os/upgrade/files/5527022/upgrade-pr127-nov11.txt)

From the above log:

```
Nov 11 15:08:39 pop-os pop-upgrade[28757]: The following additional packages will be installed:
Nov 11 15:08:39 pop-os pop-upgrade[28757]:   libpop-upgrade-gtk
Nov 11 15:08:39 pop-os pop-upgrade[28757]: The following packages will be DOWNGRADED:
Nov 11 15:08:39 pop-os pop-upgrade[28757]:   libpop-upgrade-gtk
Nov 11 15:08:39 pop-os pop-upgrade[28757]: 0 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 1 not upgraded.
Nov 11 15:08:39 pop-os pop-upgrade[28757]: 1 not fully installed or removed.
Nov 11 15:08:39 pop-os pop-upgrade[28757]: Need to get 0 B/353 kB of archives.
Nov 11 15:08:39 pop-os pop-upgrade[28757]: After this operation, 0 B of additional disk space will be used.
Nov 11 15:08:39 pop-os pop-upgrade[28757]: Do you want to continue? [Y/n] Abort.
```

That is the journalctl log, so I'm not sure why it shows an "Abort" at the yes/no prompt.

mmstick:
Easy fix. I'll have it submitted in a few minutes

jacobgkau:
Tried again, it still self-updated `pop-upgrade` to `df7c2fd` but then failed (the output was different this time.) [upgrade-pr127-nov12.txt](https://github.com/pop-os/upgrade/files/5531574/upgrade-pr127-nov12.txt)

`apt policy` shows that `libpop-upgrade-gtk` is still on `0.1.0` (the locally-built version). If I run `sudo dpkg --configure -a` at that point, it appears to finish setting up `pop-upgrade`. `sudo apt install -f` gets `libpop-upgrade-gtk` and `pop-upgrade` matching again.

mmstick:
Something's causing `pop-upgrade` to fail to complete installation, so when it restarts, it tries to upgrade again, and kills it a second time mid-release fetch.

mmstick:
I'll ensure the command outlives the daemon

jacobgkau:
Failed to build in staging and also locally:

```
   Compiling pop-upgrade v0.1.0 (/home/system76/upgrade)
error[E0599]: no method named `status` found for struct `std::process::Child` in the current scope
   --> src/daemon/mod.rs:837:59
    |
837 |             cmd.spawn().expect("apt-get failed to start").status().expect("apt-get wasn't running");
    |                                                           ^^^^^^ method not found in `std::process::Child`

```

jacobgkau:
Still failing after the daemon restart on `685135f`: [upgrade-pr127-nov13.txt](https://github.com/pop-os/upgrade/files/5539531/upgrade-pr127-nov13.txt)

mmstick:
I'll let you know when I have it working again

mmstick:
@jacobgkau We are ready.

1. Add the PPA
2. Install the debs from http://apt.pop-os.org/staging/upgrade-upgrades/pool/focal/upgrade/
3. Commence upgrade, and see that it installs the packages from the PPA, restarts the service, and continues

jacobgkau (approved):
Working as expected now (I still had to build locally to test, as installing from the staging URL gives the same version number as what's in the staging PPA.) Self-updates to the correct version and then continues the upgrade successfully. I did notice this output while watching the log when the daemon restarts:

```
Nov 17 09:02:15 pop-os systemd[1]: pop-upgrade.service: Main process exited, code=exited, status=1/FAILURE
Nov 17 09:02:15 pop-os systemd[1]: pop-upgrade.service: Failed with result 'exit-code'.
Nov 17 09:02:15 pop-os systemd[1]: pop-upgrade.service: Scheduled restart job, restart counter is at 1.
Nov 17 09:02:15 pop-os systemd[1]: Stopped Pop Upgrade Daemon.
Nov 17 09:02:15 pop-os systemd[1]: Started Pop Upgrade Daemon.
```

I'm assuming that output is expected due to the daemon restarting. Aside from the self-updating feature, upgrades with the usual problematic packages are working as expected. `pop-upgrade` is entirely up-to-date after the upgrade is complete, with no updates available and no broken dependencies or unconfigured packages. Tested with the GUI and CLI.

When navigating to the GUI, the contents of the settings page don't load until after the self-update is complete. That makes sense, because the self-update would need to happen first in order to accurately tell whether a new release is available or not. Some kind of loading indicator might be good in the future (right now the page is blank while the self-update is happening-- it looks similar to the delay on the firmware update page, so not unprecedented, but a little long.)

mmstick:
Ready to merge then?

PR #129, opened 2020-10-23
fix: Attempt systemd upgrade 10 times before giving up

mmstick:


jackpot51:
Build failures are strange. After the trubble build, I will try restarting the build server

jackpot51 (dismissed):

jackpot51:
@mmstick you could try force pushing again

jackpot51 (approved):

jacobgkau:
When upgrading an installation with `wine` installed, I still see a number of `Attempting to repair` steps, but after that, it goes back to `Installing Prerequisites` and then `Installing Updates` from 0% again. The second time through, it finishes successfully, with no manual intervention required throughout the process.

jacobgkau (approved):

PR #133, opened 2020-11-19
feat(gtk): Show loading animation while checking for updates

### LINKED ISSUE #132: (gtk) Loading indicator for OS Upgrade page in Settings
As mentioned in [this review](https://github.com/pop-os/upgrade/pull/127#pullrequestreview-532640871), the OS Upgrade page is blank while `pop-upgrade` checks for and performs a self-update.

When a self-update isn't required:

upgrade-i-screencast1

When a self-update is required:

upgrade-i-screencast2

It would be nice if there was either a loading spinner, text like "Checking for updates...", or some other indication to let users know that the system is working and not just loading slowly in those few seconds.

mmstick:
Closes #132

### DIFF
--- a/gtk/src/events/background/mod.rs
+++ b/gtk/src/events/background/mod.rs
@@ -37,6 +37,7 @@ pub fn run(
     if let Ok(ref mut client) = Client::new() {
         info!("Checking for updates to daemon");
         if client.update_and_restart().unwrap_or(false) {
+            send(UiEvent::Updating);
             let file = std::path::Path::new(pop_upgrade::RESTART_SCHEDULED);
             while file.exists() {
                 std::thread::sleep(std::time::Duration::from_secs(1));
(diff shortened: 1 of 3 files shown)


ids1024 (dismissed):

jacobgkau (changes requested):
The loading spinner looks good. However, it's just spinning indefinitely:

loading

(All `pop-upgrade` packages are up-to-date and the page is not proceeding past the spinner.)

mmstick:
That should be fixed with this change

ids1024 (dismissed):

jacobgkau:
@mmstick It looks like it's still happening.

Screenshot from 2020-11-19 13-41-33

jacobgkau (approved):
Looks good now, both when checking for a self-update and when actually performing a self-update.

- If a self-update is not available, I briefly see the spinner with no text, followed by the actual page.
- If a self-update is available, I see the spinner with no text, followed by "Updating the upgrade service", and finally "Checking for updates" before the actual page is displayed.

Release upgrades still work.

PR #135, opened 2020-11-23
fix: Correctly use shlibs for dependencies

ids1024:
Using `shlibs` to generate the run-time library dependencies is the correct method in Debian packaging conventions, and is effective at determining what dependencies and versions are needed.


mmstick (approved):

jacobgkau (approved):
Doesn't seem to affect functionality (GTK panel & release upgrade still work.)

Issue #131, opened 2020-11-06
Failing downloads do not time out


jacobgkau:
This is different from https://github.com/pop-os/upgrade/issues/107 because that issue is about `pop-upgrade` aborting on failed downloads, while this issue is about `pop-upgrade` _not_ aborting.

I am having trouble downloading packages from the Pop!_OS Launchpad PPA (speeds slow down until the download eventually fails entirely.) When this happens, `apt` will time out:

![apt-timeout](https://user-images.githubusercontent.com/7199422/98387954-51913080-200f-11eb-9395-181dedc2ecfe.png)

`pop-upgrade`, on the other hand, will hang in the fetching stage:

![pop-upgrade-lack-of-timeout](https://user-images.githubusercontent.com/7199422/98388061-7a192a80-200f-11eb-8922-2ac55f537a20.png)

Examining the complete log shows that the two packages that have started being fetched but haven't finished are `udev_246.6-1ubuntu1pop0~1603125948~20.10~d708181_amd64.deb` (line 1310/2685) and `nvidia-kernel-source-455_455.28-1pop0~1604073177~20.10~ef6315f_amd64.deb` (line 1356/2685.) Because `pop-upgrade` downloads more than one package at a time, it seems it continues fetching other packages until it reaches the end of the list, but never stops these failing downloads to retry.

Since the downloads are not going to complete, this means I have to kill `pop-upgrade` (some users would probably reboot, as GNOME Control Center is stuck open during the upgrade and there's no way to cancel/abort.)

This is happening on both `05600dc` and `3f2d44f` (so it's not currently resolved by https://github.com/pop-os/upgrade/pull/127.) Ideally, `pop-upgrade` would recognize that a download is not going to finish (like `apt` does) and would try the download again at least once, although even just exiting with an error message would at least allow me to try again manually. If it tries again automatically, it should exit eventually if the download keeps failing.

mmstick:
I can define a timeout. It wasn't possible to have a timeout in the current package-fetching code, but it's now possible in the new code using libcurl via the isahc crate.

jacobgkau:
I believe this one's still happening based on my testing. I tried disconnecting from the network while packages were downloading, and it seemed to indefinitely keep trying the downloads it was on. If there is a timeout, it's extremely high.

PR #142, opened 2021-02-24
feat: Recovery Update & Refresh

mmstick:
Requires https://github.com/pop-os/distinst/pull/251 to exist in an ISO beforehand


bflanagin:
Tests run using debug iso with pop-os/distinst#251 applied.

1. 20.10, Unencrypted drive: Success
2. 20.10, Encrypted drive: Success

jacobgkau:
upgrade-pr142-screenshot1

Looking at the message being displayed under `Recovery Partition`, it sounds a little awkward-- `You have the most current version of the recovery version`. Is it supposed to say `most current version of the recovery partition`?

WatchMkr:
And the GCC panel name adds recovery. "OS Upgrade & Recovery"

jacobgkau:
@mmstick It looks like this failed to build after the last change, it's no longer in the `rec` staging repo on Groovy or Focal. Trying to build locally gave the following error:

```
   Compiling pop-upgrade-gtk v0.1.1 (/home/system76/upgrade/gtk)
error[E0599]: no variant or associated item named `Reset` found for enum `events::OsRecoveryEvent` in the current scope
   --> gtk/src/events/mod.rs:238:34
    |
66  | pub enum OsRecoveryEvent {
    | ------------------------ variant or associated item `Reset` not found here
...
238 |                 OsRecoveryEvent::Reset => unimplemented!(),
    |                                  ^^^^^ variant or associated item not found in `events::OsRecoveryEvent`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `pop-upgrade-gtk`.

```

jacobgkau:
After installing from an older iso ([20.10 Intel iso 4](https://pop-iso.sfo2.cdn.digitaloceanspaces.com/20.10/amd64/intel/4/pop-os_20.10_amd64_intel_4.iso)), adding this branch, installing all updates, and rebooting, the GUI reports that my Recovery partition is up-to-date, even though that should not be the case:

Screenshot from 2021-03-22 15-32-25

When I run the Recovery upgrade command, the update is downloaded and the GUI updates to reflect that it's downloading:

Screenshot from 2021-03-22 15-32-43

So the check for a Recovery upgrade doesn't seem to work via the GUI. (Also, the `You have the most current version` message probably shouldn't display while the updated partition is downloading.)

mmstick:
The reason it says "most current" is because we don't urge updating the ISO unless the release API says the current release is marked as "urgent", or the recovery partition is not the same release. But an upgrade of the recovery partition can be forced from the command line.

jacobgkau:
The wording seems a little misleading, but it makes sense that it only needs to be updated once per major release.

Are we planning to mark the most recent isos with the distinst fixes as "urgent," and can we do that now so this can be tested? Currently, if I do a default encrypted install from that older iso and add this branch, I don't get the option to update the Recovery partition, but I do get the option to refresh; if I click the `Refresh` button and confirm, I'm rebooted into a Recovery partition with a distinst version that fails to handle the LUKS volume properly:

Screenshot from 2021-03-24 17-32-51

Recovering from this requires manually changing the systemd-boot default from Recovery back to the normal `pop_os-current.conf` entry.

mmstick:
Yeah, the ISO with the fix should be marked as urgent.

mmstick:
Next time ISOs are built, we can set URGENT=1 to have it marked as urgent when it's released.

jacobgkau:
We released 20.10 build #<span></span>13 this morning, and it is marked as urgent.

image

However, after doing another default encrypted install from the build #<span></span>2 iso, then adding this staging repo, I'm still not seeing an option to update the Recovery partition via the GUI:

Screenshot from 2021-03-29 09-42-11

jacobgkau:
I performed an 18.04 installation and upgraded it to 20.04. That machine _did_ show a recovery upgrade available:

recovery-upgrade-available

So the issue of not showing an available upgrade seems to only be with urgent updates for the same release (which is still an issue.) When I tried to update the recovery partition on the 18.04 -> 20.04 machine, it hung at 187 MiB:

recovery-upgrade-downloading

It eventually failed, and I saw the same error message I would see for a failed release upgrade:

recovery-upgrade-failed

I'm not sure if the generic section would be relevant for a recovery upgrade, although the actual error is present at the bottom. After dismissing that message, I saw that the download progress was still present in the label, and the release upgrade button was still greyed out:

recovery-upgrade-after-failure

Is that expected? (The download seemed to resume from where it left off when I clicked `Update` again, so if the progress is still valid after a failure, that's fine.) This time, the download finished, and I briefly saw `Verifying the fetched recovery image` followed by `Syncing recovery image to disk` before the process completed successfully:

Screenshot from 2021-03-30 08-39-18

Once it was finished, the release upgrade button became active again, and the refresh button also became active (where it was disabled while the recovery partition was out-of-date.) I upgraded that 20.04 machine to 20.10, and the recovery partition was also offering an upgrade, upgraded successfully, and was able to do a refresh install.

I then reinstalled 20.10 from an older iso (used #3 this time), was not offered a recovery upgrade, and recreated the issue where I get stuck at a blank refresh screen. So the only remaining issues seem to be:

- Not being offered a recovery upgrade for an urgent iso release.
- The error message on a failed iso download not being more tailored to the recovery upgrade (for a failed iso download, no system files would have been modified, and trying again should probably be recommended before contacting support.)
- I've also noticed that while the `Updating the upgrade service` process does work for self-updating `pop-upgrade` when accessing the GUI, any time that happens, I get `Failed to check for updates` and have to re-launch the GUI before I can use it. This one seems to be pre-existing, but thought I'd mention it too (I'll open an issue for it if none of the current PRs fix it.)

jacobgkau (changes requested):
Urgent iso detection is now working. After installing from an older iso, I'm no longer able to start a refresh install without updating the recovery partition first (which is correct behavior). However, I am now getting an error every time I try to update the recovery partition:

Screenshot from 2021-04-05 10-40-19

The error message says `failed status`, which doesn't seem correct. Every time I click `Update`, I get a line that says `Setting status to 0` in the terminal.

While addressing this, it would also be good to rebase this so it's not out-of-date with the master branch.

src/recovery/errors.rs
```diff
@@ -14,6 +14,9 @@ pub enum RecoveryError {
     #[error("failed to fetch release data from server")]
     ApiError(#[from] ApiError),
 
+    #[error("generic error")]
+    Anyhow(#[from] anyhow::Error),
```
danieleades, line 18: `anyhow` should really only be used in *binary* code, not in *library* code. Anyhow is a convenience for errors whose destiny is to be reported, rather than handled. A *library* shouldn't be making decisions about *how* it's errors are handling (that's up to the application). The library should use concrete error types (thiserror).

This is slightly more murky in this project because of the shared source directory pattern being used (lib + bin in one source directory), but it's rarely a good idea for a single build target to mix both `thiserror` and `anyhow`.

src/recovery/mod.rs
```diff
@@ -1,6 +1,7 @@
 mod errors;
 mod version;
 
+use anyhow::Context;
```
danieleades, line 4: as above. library code should use concrete error types, rather than type-erased errors.

mmstick: This project isn't a library. Useful error messages are more important in our logs.

danieleades: > This project isn't a library

in that case, why use the shared source directory? you're compiling a library target. I can think of a few good reasons to do this

- to overcome limitations that make it difficult to run integration tests against binary code
- in order to publish a subset of the codebase as a reusable library
- separation of concerns

the first doesn't apply- there are no integration tests. I don't think the second case applies, but if it did, it would be poor practice to use `thiserror`, since it should be up to the projects that *consume* the library to decide what to do with the concrete errors. Similarly for the third case- *presenting* errors would be a concern of the application target, not the binary target.

> Useful error messages are more important in our logs.

that's fine. i guess I would point out that using concrete error types in your library code doesn't preclude you from pretty printing the errors downstream in the application code (though it may well make it more verbose!). I'm *definitely* being overly pedantic here, but I guess i'm just trying to point out that you're using two, quite different error-handling strategies interchangeably in the same compilation target.

danieleades: I sound like some pedant who just wandered in off the street ;)

Feel free to go ahead and ignore all this

mmstick: > in that case, why use the shared source directory

That is because the pop-upgrade project has many components. The majority of the product is the daemon, but there's also a command line frontend, notification service, and a GTK widget to integrate into GNOME Settings. The GTK widget is its own sub-project that leans on some of the same code and types as the daemon and CLI frontend.

> I guess i'm just trying to point out that you're using two, quite different error-handling strategies interchangeably in the same compilation target.

That's what happens when you have a large project that's been developed over a decent period of time. The error handling strategy in pop-upgrade has changed as better error handling strategies in Rust have been developed. The newer code is written to take advantage of anyhow::Context, while the original code is still using thiserror. The old code isn't broken, and it hasn't been a priority to redo everything for the sake of doing away with thiserror.

danieleades: > That is because the pop-upgrade project has many components. 

why does the number of components determine whether the crate root is a `lib.rs`, a `main.rs`, or both? i'm not sure i understand the motivation for having both in your case. what value are you getting from the library target (compared to, say, a single binary target with submodules?

>The newer code is written to take advantage of anyhow::Context, while the original code is still using thiserror.

`anyhow` isn't newer. they were released at the same time, by the same author, for different use-cases.

mmstick: > anyhow isn't newer. they were released at the same time, by the same author, for different use-cases.

Feel free to go 2 years in the past of our commit history. This project began development a full year before thiserror and anyhow existed. The concept of deriving errors from enums existed before thiserror was released. The thiserror crate was a simple drop-in replacement for err-derive.

> why does the number of components determine whether the crate root is a lib.rs, a main.rs

The GTK widget generates a lib.so file to import into a C application. A lib.rs is necessary to import pop-upgrade into the GTK sub-project. Feel free to investigate our Makefile to see how everything is packaged.

danieleades: Ah that makes sense then. Thanks for the taking the time to explain, I appreciate it

jacobgkau:
Build is failing:
<details>
  <summary>Output:</summary>
  
```
   Compiling pop-upgrade v0.1.1 (/home/system76/upgrade)
error[E0599]: no method named `ok_or` found for enum `std::result::Result<(std::boxed::Box<str>, u16), anyhow::Error>` in the current scope
   --> src/recovery/mod.rs:158:58
    |
158 |                 crate::release::check::current(version_).ok_or(RecoveryError::NoBuildAvailable)?;
    |                                                          ^^^^^ method not found in `std::result::Result<(std::boxed::Box<str>, u16), anyhow::Error>`

error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> src/recovery/mod.rs:157:18
    |
157 |             let (version, build) =
    |                  ^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `std::marker::Sized` is not implemented for `str`
    = note: all local variables must have a statically known size
    = help: unsized locals are gated as an unstable feature

error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> src/recovery/mod.rs:158:17
    |
158 |                 crate::release::check::current(version_).ok_or(RecoveryError::NoBuildAvailable)?;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `std::marker::Sized` is not implemented for `str`
    = note: only the last element of a tuple may have a dynamically sized type

error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> src/recovery/mod.rs:163:17
    |
163 |                 info!("recovery partition is already upgraded to {}b{}", version, build);
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `std::marker::Sized` is not implemented for `str`
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 4 previous errors

Some errors have detailed explanations: E0277, E0599.
For more information about an error, try `rustc --explain E0277`.
error: could not compile `pop-upgrade`.
```
</details>

jacobgkau:
Urgent iso updates are now working.

On a 20.10 machine, the `Refresh` button is no longer booting into recovery mode. The system reboots back into the normal OS. Looking in `/boot/efi/loader/loader.conf`, it doesn't look like the default boot option is being changed. I performed a fresh install from an older 20.10 iso and this still happened; however, it's not happening on a 20.04 machine.

Also, similar to https://github.com/pop-os/upgrade/issues/131, if a networking change occurs (such as unplugging Ethernet, or just plugging in Ethernet while downloading via WiFi), the iso download hangs and never starts again unless I close and re-open GNOME Control Center (and even then, it sometimes hangs when I try downloading again.)

Finally, the download failure error message seems to still be non-descript (`failed status`). I saw this on the 20.04 machine without trying to invoke any errors (the download did work when I retried a few minutes later):

Screenshot from 2021-04-09 11-51-16

To clarify the last two points, that error message doesn't show up when I unplugged Ethernet, it only showed up when the download failed for some other reason.

mmstick:
The error happens because the API changed has changed, but the old client is still loaded in memory in the GTK widget, so re-opening the page is necessary to get the new client.

jacobgkau:
I had a failed upgrade on this branch when going from Bionic to Focal with a number of non-default packages installed. I was rebooted back into Bionic with my sources set to Focal (but `/etc/os-release` still saying Bionic and Bionic packages still installed), no `upgrade.log` in `/var/log`, missing icons, and a dependency issue requiring `apt install -f` to be run after I reverted the sources. Currently trying to recreate and confirm if it's a regression or not.

mmstick:
Logs from the daemon would be helpful to check if the APIs are being called correctly

jacobgkau:
I recreated the failure by installing the same set of packages that were present before the first failure: [installed-bionic.txt](https://github.com/pop-os/upgrade/files/6423460/installed-bionic.txt)
A script to install those packages: [installed-bionic-oneline.txt](https://github.com/pop-os/upgrade/files/6423461/installed-bionic-oneline.txt)

Daemon logs from after the first failure: [pop-upgrade-journal.txt](https://github.com/pop-os/upgrade/files/6423463/pop-upgrade-journal.txt)
Daemon logs from before the second failure: [pop-upgrade-before-failure2.txt](https://github.com/pop-os/upgrade/files/6423464/pop-upgrade-before-failure2.txt)

This screenshot demonstrates some of the missing icons after the failure (including all icons in the top right of the screen):
Screenshot from 2021-05-04 11-31-36

I tried a third time without adding this branch to the machine, and the issue still occurred, so it does look like this isn't a regression.

mmstick:
Will fix the issue so that we have one update containing all the fixes we need.

jacobgkau (changes requested):
It seems some of the internationalization is not applying correctly. Refresh copy is missing:

Screenshot from 2021-05-24 11-50-49

Download label is not formatted properly:

Screenshot from 2021-05-24 12-10-09

jacobgkau:
The download label is fixed, but the refresh copy is still cut off:

upgrade-pr142-refreshcopy

Do i18n strings support multiple lines? It seems like newline characters might be needed instead of the way it's [currently formatted](https://github.com/pop-os/upgrade/blob/f267a64bff1043aa18d3ec11eddaa41a25742e1d/i18n/en/pop_upgrade_gtk.ftl#L17).

mmstick:
It does, but the empty lines needs to have a space preserved, which my editor has stripped on save

jacobgkau:
Not fixed yet on `c43b1a7`.

mmstick:
Confirmed that I've fixed it. It seems that a line cannot start with `*` unless it has been escaped

jacobgkau:
Yep, it's displaying now:

Screenshot from 2021-05-25 07-33-47

For the fourth bullet point:

```
All files outside of the /home directory in the OS partition will be lost
```

This makes it sound like files on non-OS partitions (e.g. extra drives) would be removed, when they actually aren't. Can we change the wording to this?

```
All files in the OS partition outside of the /home directory will be lost
```

Also, Flatpak applications installed via the Pop!_Shop (which uses user mode by default) are not removed, so this bullet point is not accurate:

```
All applications installed by the user will be removed
```

Does that need to be clarified? (Might need input from @maria-komarova or others about the messaging here?)

mmstick:
Sure. Things have changed and it'd be good to mention that.

maria-komarova:
> All files in the OS partition outside of the /home directory will be lost

Makes perfect sense. 

For the second one, we could probably do something like 
> All applications installed by the user will be removed with the exception of Flatpak packages installed in Pop!_Shop

But I wonder if that makes it unnecessarily complex. People might not remember if they installed some Flatpak application through Pop!_Shop or through terminal. So maybe better say all will be removed? I am torn.

mmstick:
Could say "System applications", since system-installed Flatpaks are also lost, but the locally-installed Flatpaks are retained. We don't yet support installing Flatpaks system-wide in the shop, but that feature is coming soon.

maria-komarova:
Just to clarify, we are installing Flatpaks per user in the Pop!_Shop currently?

maria-komarova:
Is this what you mean @mmstick? 
> All system applications installed by the user will be removed

Would that be clearer I wonder?

jacobgkau (changes requested):
After upgrading from 18.04 to 20.04, I am unable to dismiss 20.10 notifications through the GUI:

dismiss

The file at `/usr/lib/pop-upgrade/dismissed` is not created, and the button does not grey out.

Running `pop-upgrade release dismiss` does work, and after doing that and restarting GNOME Control Center, the GUI button greys out. Confirmed the GUI button works on current stable, so this is a regression.

jacobgkau:
On `4ec2f45`, the button now works for adding the `dismissed` file, but the button doesn't grey out until I restart GNOME Control Center. On stable, it greys out immediately after being clicked.

Also, looks like `4ec2f45` failed to build for Bionic for some reason?

jacobgkau:
One other thing, if I make sure the `dismissed` file has been removed, I get a notification for version 20.10 when logging in with stable installed, while this branch doesn't display a notification. Running `pop-upgrade release check` in a terminal tells me a new version is available, but also doesn't display a notification. (`e5fa1aa` hasn't finished building yet, so I haven't tried it yet, but it doesn't seem like displaying of the notification should be related since all that's changing is the dismissal button.)

jacobgkau:
The latest version, `e5fa1aa`, still does not grey out the `Dismiss` button when it's clicked (just keeping track of that.)

jacobgkau:
`e5fa1aa` does grey out the `Dismiss` button when it's clicked, but still doesn't display notifications when not dismissed.

jacobgkau (approved):
Notifications are working again. The dismiss button is still working. The revised Refresh copy is still showing properly. Refreshing still works.

- 18.04 installation + updates + Kdenlive, VLC, Chromium, Spotify -> 20.04 + transition Spotify, install Steam -> 20.10: successful.
- 18.04 -> 20.04 using the earlier problematic package list still works.
- 20.04 + WINE + GIMP + Inkscape -> 20.10 + Meld + MPV -> 21.04: successful.

All regressions have been taken care of. I think we should go ahead and merge this, and further improvements can come in later PRs.

mmstick:
Awesome. This is a huge improvement.

PR #144, opened 2021-03-08
Clippy linting

danieleades:
I've fixed a couple of little nits, and also inadvertently fixed some formatting (by applying `cargo +nightly fmt`)

i think perhaps the nits would have been denied by clippy, and the formatting is easy enough to enforce in CI.

These could both be easily added as either github actions, or [pre-commit hooks](https://github.com/doublify/pre-commit-rust)

I note also that the 'expected checks' don't appear to be running on this pull request.


jacobgkau (changes requested):
This failed to compile on the build server:

```
Compiling pop-upgrade v0.1.1 (/<<PKGBUILDDIR>>)
error[E0308]: mismatched types
   --> src/daemon/mod.rs:586:59
    |
586 |         let apt_uris = crate::fetch::apt::fetch_uris(Some(&borrows)).await?;
    |                                                           ^^^^^^^^ expected slice `[&str]`, found struct `std::vec::Vec`
    |
    = note: expected reference `&[&str]`
               found reference `&std::vec::Vec<&std::string::String>`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `pop-upgrade`.
```

I'm getting the same `mismatched types` error when trying to build locally.

danieleades:
i've reverted this change.

There's some gymnastics going on with these `&[&str]` arguments

jacobgkau (dismissed):
Upgrades still seem to work as expected from 18.04 to 20.04 and 20.04 to 20.10. `release repair` command is also still working.

src/cli/mod.rs
```diff
@@ -470,28 +470,23 @@ impl Client {
...
-                    ),
+                } else {
+                    error!("major ({}) and minor({}) version failed to parse as u8", major, minor)
                 }
+            } else {
+                error!("version {} is invalid", next)
             }
-            None => error!("version {} is invalid", next),
-        },
+        }
```
danieleades, line 489: using `if let` blocks removes two levels of nesting here.

that said, i think this should be refactored further (but perhaps not in this pull request). it's not the most readable function.

danieleades:
i've added a CI target for clippy.

this could be further improved by using the [clippy-check](https://github.com/actions-rs/clippy-check) action, which is able to also comment on pull requests. that would require setting up a github token.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: install system deps
+        run: sudo apt install libdbus-1-dev libparted-dev
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          components: clippy
+      - uses: actions-rs/cargo@v1
```
crawfxrd, line 34: use [`clippy-check@v1`](https://github.com/actions-rs/clippy-check)

.github/workflows/CI.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: install system deps
+        run: sudo apt install libdbus-1-dev libparted-dev
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+      - uses: actions-rs/cargo@v1
+        with:
+          command: check
```
crawfxrd, line 21: This is unnecessary as clippy runs check itself.

danieleades: this is true. i've pulled this from a larger CI pipeline from another project. the 'check' target is useful for 'failing fast'. the job will typically complete much faster than the 'clippy' target (especially on large codebases)

if 'failing' fast isn't useful, i can strip out the 'check' target

.github/workflows/CI.yml
danieleades: see my [comment above](https://github.com/pop-os/upgrade/pull/144#issuecomment-811702277).

clippy-check expects a github token to be set up, i believe. but it *is* my preferred option

danieleades: ... which i suddenly see [is created by default](https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret).

i'll update the pull request

danieleades:
job is passing - https://github.com/danieleades/upgrade/actions/runs/709030800

danieleades:
rebased on master

src/release/repos.rs
```diff
@@ -287,11 +287,11 @@ X-Repolib-Default-Mirror: http://us.archive.ubuntu.com/ubuntu/
...
-    format!(
-        r#"## This file is deprecated in Pop!_OS.
-## See `man deb822` and /etc/apt/sources.list.d/system.sources.
-"#
-    )
+    r###"
+    ## This file is deprecated in Pop!_OS. 
+    ## See `man deb822` and /etc/apt/sources.list.d/system.sources.
+    "###
+    .to_string()
```
crawfxrd, line 294: Trying this on playground, the leading indent shows up in the output.

danieleades: rats. that's embarrassing. try now!

danieleades:
most recent CI results can be checked here - https://github.com/danieleades/upgrade/runs/2354564628

it's failing because i added GTK, but haven't added the build dependencies yet...

will update this evening

danieleades:
@crawfxrd are you happy for me to squash these edits into one/a few commit(s)?

while there are a lot of code changes, they're largely cosmetic (modulo any mistakes!).

there's not really any rhyme or reason to the current commits (i had assumed these would be squash merged)

crawfxrd:
Yeah, I think could be squashed to 2 commits: One for the CI, and one for applying the clippy fixes.

danieleades:
> Yeah, I think could be squashed to 2 commits: One for the CI, and one for applying the clippy fixes.

Great minds, etc

danieleades:
action is passing - https://github.com/danieleades/upgrade/runs/2368389890
commits are rebased and tidied.

ready for review

danieleades:
quite a lot of dead code in the `gtk` package - https://github.com/danieleades/upgrade/runs/2368405115?check_suite_focus=true

gtk/src/gtk_utils.rs
```diff
@@ -1,7 +1,7 @@
 use gtk::prelude::*;
 
 pub fn scale_label(label: &gtk::Label, scale: f64) {
-    let list = label.get_attributes().unwrap_or_else(|| pango::AttrList::new());
+    let list = label.get_attributes().unwrap_or_else(pango::AttrList::new);
```
crawfxrd, line 4: Does `AttrList` implement `Default`? If so, use `unwrap_or_default()`.

danieleades: good catch. done.

gtk/src/widgets/dialogs/mod.rs
```diff
@@ -23,7 +23,7 @@ fn dialog_template<F: FnOnce(&gtk::Box)>(
         .accept_focus(true)
         .deletable(true)
         .destroy_with_parent(true)
-        .use_header_bar(1i32)
+        .use_header_bar(1_i32)
```
crawfxrd, line 26: This is a pedantic lint, which isn't enabled and so not required in this change.

danieleades: actually there's a few refactors in this pull request that aren't strictly required to satisfy the `clippy::all` lint. is that an issue?

crawfxrd: I think if they're being changed then they should have the corresponding lint enabled. The whole set may be a lot of noise, so it may be better to pick a subset to enforce.

danieleades: done.

crawfxrd: The rule was added without fixing it; clippy will now fail. Only 3 instances were changed, so it's easier to just drop this lint.

mmstick:
We can do this later

PR #149, opened 2021-03-27
use 2018 edition idioms

danieleades:
this refactors the crate to use 2018 edition idioms

- explicit macro imports
- 2018-style modules
- apply formatting (`cargo +nightly fmt`)


danieleades:
this would add 2018 edition idioms just in time for the 2021 edition...

the only place this is likely to bite even slightly is the 'extern crate' macro imports. there's a decent chance this style of import will become a compiler warning in 2021 edition (see https://github.com/rust-lang/rust/issues/80165). Not a hard error, mind.

Although keeping the 2015 idioms isn't actually going to cause any real problems, i'm inclined to believe the 2018 idioms were introduced for good reasons.

- i'm a big fan of explicit macro imports. i think implicit global imports are a bit 'magic', and it's weird that they're treated differently to every other first-class object
- the 2018-style module structure is a big win if you've ever had multiple `mod.rs` files open in an IDE at once. I find that the file paths are less ambiguous, and i'm immediately more 'context-aware' working on code. Makes me less likely to 'stutter' in the fully-qualified name of an object (ie. `recovery::version::RecoveryVersionError`, which could just as easily be `recovery::version::Error`)

src/daemon/methods.rs
```diff
@@ -9,6 +9,7 @@ use dbus::{
 };
 
 use crate::misc::format_error;
+use num_derive::FromPrimitive;
 use num_traits::FromPrimitive;
```
crawfxrd, line 13: Are both imports required?

crawfxrd:
The commit dropping the `cli2` drirectory should be applied as a fixup to the commit that introduced it.

I think the explicit imports should be collapsed into one commit, since they're doing the same thing.

danieleades:
> The commit dropping the `cli2` directory should be applied as a fixup to the commit that introduced it.
> 
> I think the explicit imports should be collapsed into one commit, since they're doing the same thing.

I had assumed this would all be applied as a single squash merge

crawfxrd:
The repo appears to be configured for squash merging by default (don't know why), but the preferred method is rebase merging.

danieleades:
> The repo appears to be configured for squash merging by default (don't know why), but the preferred method is rebase merging.

i'm tempted to rebase this as a single commit (which *is* a squash merge) unless you particularly want these commits separated?

src/daemon/methods.rs
danieleades: yep. though i agree it looks weird. also surprising that they don't collide (they don't)

crawfxrd:
I think how you did it in system76-power is good. One commit for moving files for 2018-style crates, and one commit for explicit imports. There's some rustfmt changes mixed in, which could also be a separate commit.

src/daemon/methods.rs
ids1024: > also surprising that they don't collide

Macros are in a different namespace from types, which I guess is both a legacy from the pre-2018 way of importing macros, and because a derive macro is typically the same name as the trait it derives.

danieleades:
i've rebased and tidied the commits.

i've left the formatting, for the simple reason that it would be painful to change. #155 can prevent this kind of noise happening again. Don't really mind which gets merged first, i can rebase either pull request on the other without too much pain.

crawfxrd (approved):

bflanagin (approved):
No regressions found.

Issue #141, opened 2021-02-23
"pop-upgrade recovery default-boot" panics


j-tai:
This command causes a panic:

```
$ RUST_BACKTRACE=1 pop-upgrade recovery default-boot
checking if pop-upgrade requires an update
thread 'main' panicked at 'internal error: entered unreachable code', src/cli/mod.rs:73:18
stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: __libc_start_main
  15: <unknown>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

bflanagin:
Confirmed in Pop!_OS 20.10 
 
pop-upgrade:
  Installed: 0.1.1~1608137131~20.10~d3bfb98
  Candidate: 0.1.1~1608137131~20.10~d3bfb98
  Version table:
 *** 0.1.1~1608137131~20.10~d3bfb98 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main amd64 Packages
        100 /var/lib/dpkg/status

danieleades:
Looking at the code, while the command exists, it's not actually implemented.

j-tai:
Looks like this is fixed.

PR #150, opened 2021-03-28
use declarative syntax for defining CLI

danieleades:
use Clap's strongly-typed, declarative syntax for defining the CLI (instead of the fluent, imperative style).

I've used a pinned version of the clap 3.0 beta here, but this could easily be changed to StructOpt if is this is a concern.

advantages-

- deserialisation is handled implicitly by clap
- no need to manually align CLI definition with CLI parsing
- recursive structure provides better encapsulation of subcommands
- args can be documented with rustdoc (if anyone cares about that?)

**This will require careful review and testing, as these are broad changes that i've not been able to test locally**


src/cli/recovery.rs
```diff
@@ -0,0 +1,155 @@
...
+
+const RESULT_STR: &str = "Recovery upgrade status";
+const RESULT_SUCCESS: &str = "recovery partition refueled and ready to go";
+const RESULT_ERROR: &str = "recovery upgrade aborted";
+
+/// tools for managing the recovery partition
+#[derive(Debug, Clap)]
+pub enum Command {
+    /// set the recovery partition as the default boot target
+    DefaultBoot {
```
danieleades, line 15: this doesn't seem to be implemented. can it be removed?

src/cli/release/repair.rs
```diff
@@ -0,0 +1,14 @@
+use clap::Clap;
+use pop_upgrade::client::{Client, Error as ClientError};
+
+/// search for issues in the system, and repair them
+#[derive(Debug, Clap)]
+pub struct Command {
+    /// Attempt to upgrade to the next release, even if it is not released
+    #[clap(short, long)]
+    force_next: bool,
```
danieleades, line 9: this arg doesn't seem to be used??

danieleades:
I also removed the qualified name 'stuttering' and glob imports in the `cli::colors` module

ids1024:
I've been thinking about using this for Pop OS projects, since the old clap API can lead to issues that can't be caught at compile time.

Any idea when a stable clap 3.0 is likely to be released? Or if there are any known issues with it blocking release? I know 3.0 has been in beta for a while, but I haven't been following too closely.

I assume pulling in a proc macro here also increasing compile time somewhat. Though it's probably worthwhile given the advantages.

danieleades:
> I've been thinking about using this for Pop OS projects, since the old clap API can lead to issues that can't be caught at compile time.
> 
> Any idea when a stable clap 3.0 is likely to be released? Or if there are any known issues with it blocking release? I know 3.0 has been in beta for a while, but I haven't been following too closely.
> 
> I assume pulling in a proc macro here also increasing compile time somewhat. Though it's probably worthwhile given the advantages.

Clap 3.0 has been in beta for quite a long time. Looking at the source, it looks like their documentation is lagging a little, so it that could be the blocker. the code itself has been stabilised for years as [`StructOpt`](https://docs.rs/structopt/0.3.21/structopt/). Structopt is a wrapper around the original clap api, and now structopt is being absorbed *back* into the main clap repo as Clap 3.0.

i've pinned the dependency, which i think mitigates any risk associated with instability. but like i say, can switch to structopt (with a find and replace) if it's a concern.

You might get a negligible increase in compile time due to the proc macro, but not so you'd ever notice.

as you can probably see, the unused arguments and unimplemented subcommands in the original implementation stand out like a sore thumb with the derive-based interface

mmstick:
May want to hold off until the recovery PR is merged

danieleades:
> May want to hold off until the recovery PR is merged

yep, fair enough.

I took a quick look at #142 (i assume that's the one you mean?) and the overlap is fairly minimal, so it shouldn't be too hard to rebase either one on the other.

I notice you've switched to `anyhow::Error` in the binary. makes sense.

I would caution against replacing all of the `client::Error`s out of hand. no point erasing the error type until the higher-level calling code where you start to coalesce multiple error types.

Also, I notice you have an `anyhow::Error` in the library code-

//`src/recovery/errors.rs`//
```rust
    #[error("generic error")]
    Anyhow(#[from] anyhow::Error),
```
is that intentional? Seems to be a mixture of `thiserror` and `anyhow`. Normally I would only use `anyhow` in application code, and never in library code.
Can be hard to tell the two apart in a shared source directory, I might be getting mixed up.

danieleades:
i want to stress that this will need to be gone over with a microscope, since i haven't tested this pull request *at all* locally!

I have to tell you, it feels very satisfying to writing pull requests for Pop OS components from my Pop OS laptop. big fan.

src/cli/recovery.rs
```diff
@@ -0,0 +1,136 @@
...
+        reboot: bool,
+    },
+
+    Upgrade(Upgrade),
+}
+
+impl Command {
+    pub fn run(&self, client: &Client) -> Result<(), Error> {
+        match self {
+            Self::DefaultBoot { reboot: _ } => unimplemented!(),
```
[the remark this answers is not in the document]
danieleades: see #141

danieleades:
closing in favour of #249

PR #151, opened 2021-03-29
Enable development upgrades to 21.04

jackpot51:
This _should_ allow upgrades to 21.04 using the --force-next flag. I have not tested it yet, will do so after it builds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2034,7 +2034,7 @@ dependencies = [
 
 [[package]]
 name = "pop-upgrade"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "anyhow",
  "apt-cmd",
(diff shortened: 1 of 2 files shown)


mmstick (approved):

bflanagin:
Butting in on this ticket. It works on my machine. 

Without --force-next:
```
$pop-upgrade release upgrade

Current Release: 20.10
Upgrading to: 21.04
New version available: false
no release available to upgrade to
```
With --force-next:
upgrade begins, and finishes correctly

bflanagin:
Found one issue that might not be a huge deal. 
If you run `pop-upgrade release upgrade --force-next` when you are already running the latest test version of pop (21.04) upgrade will attempt to download the next version,in this case 21.10, and hang.

jackpot51:
@bflanagin what is the output when it hangs?

jackpot51:
@bflanagin I think that is expected right now

PR #154, opened 2021-03-30
Fix user prompt

### LINKED ISSUE #115: Pressing Enter at "Try again? y/N" does not select "N" as the default option
In the below screenshot, the capital "N" would typically indicate that it's the default option if the user presses Enter without typing any characters:

2004-to-2010-force

However, as shown, the n (or N) option is not selected by default.

danieleades:
closes #115


src/cli/mod.rs
```diff
@@ -584,45 +584,42 @@ fn log_result(
...
+
+        let mut buf = String::default();
+        std_in.read_line(&mut buf)?;
+        Ok(buf)
     }
 
-    let mut display_prompt = move || -> io::Result<Answer> {
-        answer.clear();
-
-        stdout.write_all(message.as_bytes())?;
```
danieleades, line 604: i've dropped the `bytes` handling. what is going to go wrong because i haven't done that?

danieleades:
added some unit tests.

to run- `cargo test --bin pop-upgrade`

danieleades:
rebased and tidied commits

danieleades:
Hey @mmstick this is broken on master. Are you perhaps not a fan of the approach in this pull request? Happy to respond to any feedback

mmstick:
This is in a development branch now

PR #155, opened 2021-03-31
Apply and enforce formatting

### LINKED ISSUE #152: Rustfmt is not being applied/checked
i've submitted a couple of pull requests now and i notice that rustfmt hasn't been applied to this crate. That is, if i run rustfmt according to the 'contributing' guide, then it makes changes to the formatting.

while not a big deal, it could lead to a few mildly painful rebases. It's also really easy to catch.

I'm happy to submit a pull request for this?

could catch it using a github action, and/or a pre-commit hook.

thoughts?

danieleades:
apply formatting, and add a CI action to check formatting

closes #152


danieleades:
rebased on master

danieleades:
rebase and tidy commits

danieleades:
github action run results can be checked here - https://github.com/danieleades/upgrade/pull/3/checks?check_run_id=2360001747

src/release/check.rs
```diff
@@ -125,12 +125,12 @@ fn next_(
             next = "21.04";
 
             ReleaseStatus {
-                build:   if development { release_check(next) } else { BuildStatus::Blacklisted },
+                build: if development { release_check(next) } else { BuildStatus::Blacklisted },
                 current: "20.10",
-                is_lts:  false,
+                is_lts: false,
```
crawfxrd, line 130: These don't seem right: it's *removing* the field alignment, whereas everything else *adds* field alignment. What's causing fmt to do that?

danieleades: Good catch. Rustfmt bug? Will do some digging

danieleades: if the difference in field widths is too great, this formatting rule doesn't apply. this threshold is configured to 30 characters. the max diff of field widths in this struct is greater than 30, so this is correct

the rule itself is unstable

bflanagin (approved):
No regressions detected.

crawfxrd (approved):

mmstick:
Formatting changes will destroy mergeability of development branch. Had to undo the recently-merged commits.

danieleades:
@mmstick I take it you don't want to persue enforcing the formatting policy in CI?

mmstick:
Maybe when it's not actively in development

danieleades:
> Maybe when it's not actively in development

Fair enough. Though I would point out that a pull request to add formatting is, by it's very nature, very easy to rebase.

PR #162, opened 2021-05-19
Add prerequisite packages, check for presence before upgrading

jacobgkau:
This resolves the upgrade failure caused by the list of packages I provided in [this comment](https://github.com/pop-os/upgrade/pull/142#issuecomment-832205986). (I diagnosed this by running the `upgrade.sh` script manually and observing `not yet configured` messages when the failure occurred.)

Expanded the existing `PREINST` variable with more packages that need to be installed, and tweaked the function using that variable to sequentially check if each package is present and, if so, upgrade it. (This was necessary because `libmount1:i386` is not present by default on 18.04.)

So far, I've tested that upgrades from a default 18.04 installation (both NVIDIA and Intel) and of an installation with the problematic package list all succeed. I'm still doing some final testing to make sure all three of these packages are really needed (looks like the last one may not need to be there), and also need to regression test for 20.04 -> 20.10 (although heavier testing can wait until this is merged into the `rec` branch.)


mmstick:
I'll merge to rec when you're ready

jacobgkau:
Confirmed that `libpcre3:amd64` did not need to be added; upgrades succeed with only `libc6:i386` and `libmount1:i386` added to the list. Also confirmed a 20.04 -> 20.10 upgrade succeeded.

I did have an issue where `libc++1` and `libc++abi1` were out of date and being kept back after the 20.04 -> 20.10 upgrade. Running an `apt full-upgrade` removed `libc++1` and `libc++1-10` along with `discord`. Installing `discord` again pulled in `libc++1` and `libc++1-11` as dependencies, and removed `libc++abi1-10`. I confirmed this also happened with the `rec` branch and is not a regression from this change.

I also found that in a virtual machine running with BIOS boot instead of EFI, after upgrading from 18.04 to 20.04, when I went to upgrade from 20.04 to 20.10 immediately after, the upgrade GUI hung while `pop-upgrade` was apparently being asked an interactive question about `grub-pc`:

Screenshot from 2021-05-20 11-43-44

I also confirmed that this one happens on `rec` and is not a regression from this change. I think this branch is good to merge into `rec`, I'll check these problems out compared to `master` when testing #142.

PR #171, opened 2021-06-06
i18n(bn): Add Bangla translation

Oymate:


jacobgkau (approved):
Thank you for the translation!

Translations don't seem to be working for the pop-upgrade Settings panel right now, but this `.ftl` file looks valid. @mmstick If it would help to have a non-English translation present for testing why they're not working, then we can go ahead and merge this one.

mmstick:
If an English translations is working, these should be working as well. If not, it's defaulting to the `en`` locale for some reason. Strange.

PR #172, opened 2021-06-06
i18n(sr): Add Serbian (Latin) translation

AAnzel:


jacobgkau:
Hmm, this doesn't seem to be working:

Снимак екрана из „2021-06-07 09-04-23“

Snimak ekrana iz „2021-06-07 09-06-05“

#171 doesn't work, either. @mmstick, these are the first non-English translations we've had for pop-upgrade, any idea why they aren't working?

PR #174, opened 2021-06-15
pt_BR translation

Canuto-IX:
This is my first pull request, I hope it helps.

### DIFF
--- a/i18n/pt_BR/pop_upgrade_gtk.ftl
+++ b/i18n/pt_BR/pop_upgrade_gtk.ftl
@@ -1,0 +1,91 @@
+-os = Pop!_OS
+
+battery-notice = <b>Conecte na tomada</b> antes de começar.
+
+button-cancel = Cancelar
+button-dismiss = Recusar
+button-download = Download
+button-perform-refresh = Reiniciar e limpar
+button-perform-upgrade = Reiniciar e Atualizar
+button-refresh = Limpar
+button-update = Atualizar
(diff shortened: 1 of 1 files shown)


mmstick:
How does it compare to #173 ?

CanutoIX:
> How does it compare to #173 ?

I've translated some technical terms and some words to make them more consistent with the rest of the system.

jacobgkau:
Like the other translation pull requests currently open, this doesn't seem to be working:

Captura de tela de 2021-06-17 07-17-31

Everything looks correct with the `.ftl` file, but we need to figure out why translations aren't applying for the pop-upgrade Settings page.

Also, not knowing Portuguese, I'm not sure how to make the call between #173 and #174. Using Google Translate as a guide, this one seems to have more literal translations for some strings (such as `Os novos recursos incluem:` instead of `Novos recursos inclusos:`), but less literal translations for others (such as `Procurando por atualizacões` instead of `Verificando por atualizações`.) #173 refers to a Refresh Install as a "repair system installation," while this one refers to it as a "clean installation," and while I can see the comparison with the "clean boot" feature in Windows, the phrase "clean installation" tends to have a different connotation (of a completely new install, as opposed to a refresh install.) But of course I don't know if that phrase carries over across languages, either.

Because #173 came in first, I'm inclined to say it would be better to either make a commit or leave a review on #173 with the necessary corrections (and explanations if possible), or merge that one first and then rebase this one to make corrections.

PR #175, opened 2021-06-24
fix(ffi): Initialize fluent locale on widget create

mmstick:
This will fix the locale defaulting to `en` in the widget from C.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2416,6 +2416,8 @@ dependencies = [
  "glib",
  "gtk",
  "gtk-sys",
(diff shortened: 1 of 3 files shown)


jacobgkau (approved):
Serbian and Bangladesh translations are now working, and English is still working. Portuguese (Brazil) still isn't, but that appears to be specific to that translation (I'm getting extra warnings in the terminal when running with that language), and we can fix it in https://github.com/pop-os/upgrade/pull/174 since that's open anyway.

PR #178, opened 2021-06-28
feat: Add cosmic favorites on upgrade

ids1024:
Without this, the launcher, workspaces, and applications buttons won't appear in the dock on upgraded systems (without manually adding them).

Haven't tested this in an actual upgrade yet, so still need to see if it works properly. Dconf might need a user dbus session...

### DIFF
--- a/src/favorites.rs
+++ b/src/favorites.rs
@@ -1,0 +1,66 @@
+use as_result::MapResult;
+use std::{
+    process::Command,
+    str,
+};
+
+const SETTINGS_JS_SCRIPT: &str = "
+const Gio = imports.gi.Gio;
+
+const COSMIC_FAVORITES = [
+    'pop-cosmic-launcher.desktop',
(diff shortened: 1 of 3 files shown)


mmstick (approved):

jacobgkau:
Can this be rebased now that the extension branch has been merged? (If we're supposed to be looking at it, that is.)

jackpot51:
Don't rebase yet! We are still working through builds

jacobgkau:
This does not seem to work for a user account that had its favorites manually modified (icons added/removed) before upgrading.

jackpot51:
@ids1024 it might be better to use https://github.com/pop-os/default-settings/blob/master_focal/usr/bin/pop-app-folders to apply this change

ids1024:
I didn't know about that script, but this still doesn't seem to work, so I guess that might be better.

jackpot51:
Let's try it there before doing anything further in `pop-upgrade`

PR #183, opened 2021-06-30
feat: Replace do-release-upgrade

### LINKED ISSUE #177: Conflict with do-release-upgrade
Users sometimes run into problems after running Ubuntu's `do-release-upgrade` script, which we do not test or support. `do-release-upgrade` is provided by `ubuntu-release-upgrader-core`, which is a dependency of `update-manager-core`, which is a recommends of `ubuntu-standard`.

`pop-upgrade` should ensure that `do-release-upgrade` is not present on the system, perhaps by conflicting with `update-manager-core` or conflicting with/providing `ubuntu-release-upgrader-core`.

mmstick:
Closes #177

### DIFF
--- a/data/do-release-upgrade
+++ b/data/do-release-upgrade
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+echo "\`do-release-upgrade\` is not supported on Pop!_OS."
+echo "Use \`pop-upgrade release upgrade\` instead."
+echo ""
+echo "For more information, visit: https://support.system76.com/articles/upgrade-pop"
+
--- a/debian/control
+++ b/debian/control
@@ -39,3 +39,21 @@ Depends:
   ${misc:Depends}
 Description: The C headers required for linking the libpop-upgrade-gtk widget
  The C headers required for linking the libpop-upgrade-gtk widget
+ 
+Package: ubuntu-release-upgrader-core
+Architecture: amd64
+Description: Replaces do-release-upgrade with a help message.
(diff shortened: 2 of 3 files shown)


ids1024:
This would prevent `pop-upgrade` from updating without an `apt full-upgrade`, right?

Another solution would be to package our own version of `ubuntu-release-upgrader-core` with a script saying to use `pop-upgrade` instead.

mmstick:
Similar to nvidia-graphics-driver updates, yes.

bflanagin:
It looks like the fix would require users to run `dist-upgrade` on their systems. Is there a way to avoid this so a novice user isn't greeted with an error message when installing via pop-shop?

jacobgkau:
Re: `nvidia-graphics-driver` updates, our NVIDIA driver updates don't require `dist-upgrade`/`full-upgrade`, we specifically test to make sure they install with a regular `upgrade`.

mmstick:
Depends on the update. Any update that deprecates and transitions to a newer set of packages requires full-upgrade.

jacobgkau:
> Software centers perform a full-upgrade by default.

Are you sure? When we [renamed the COSMIC Dock extension](https://github.com/pop-os/cosmic-dock/pull/20) during the beta, users got error messages through the Pop!_Shop and had to run `sudo apt full-upgrade` manually in a terminal, we know that because they kept asking about it in the Mattermost chat.

PR #185, opened 2021-06-30
fix: Update systemd-boot-conf dependency to fix error about missing linux field

### LINKED ISSUE #184: RefreshOS fails if there are systemd-boot entries that just launch efi files
image
image

mmstick:
Closes #184

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3149,10 +3149,11 @@ dependencies = [
 
 [[package]]
 name = "systemd-boot-conf"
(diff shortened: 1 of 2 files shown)


leviport (approved):
Before adding this patch, I added an EFI shell to systemd-boot and saw a 20.10 to 21.04 upgrade fail. After adding this patch, the upgrade was successful.

crawfxrd (approved):

PR #190, opened 2021-07-09
fix: Use dbus-rs backend of notify-rust to fix busy loop issue

ids1024:
In zbus, this is fixed upstream in https://gitlab.freedesktop.org/dbus/zbus/-/merge_requests/254, but that is not in a stable release, and the betas including it have a semver bump and higher Rustc requirement.

Patching zbus would also work, but since the `dbus` create is already a dependency, this seems reasonable if it works alright.

Fix for https://github.com/pop-os/pop/issues/1792.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -721,17 +721,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derivative"
-version = "2.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jackpot51 (approved):

jacobgkau (approved):
Confirmed this fixes the high CPU issue on 21.04, 20.10, and 20.04 LTS. I still receive notifications from both `pop-upgrade` and `firmware-manager` when I'm supposed to.

PR #197, opened 2021-07-21
i18n(de): Add german translation

Salatwurzel:
Here is my translation. I double checked, everything should be good.


jacobgkau (approved):
Thank you for the translation! This displays properly in a user account set to German.

mmstick (approved):

PR #204, opened 2021-08-08
i18n(ja) : add Japanese translation

tokida:
Add Japanese translation


jacobgkau (approved):
Thank you for the translation! CI is passing on all current releases, confirmed this is working on a 21.04 machine.

jackpot51 (approved):

PR #206, opened 2021-08-09
i18n(nb_NO): add Norwegian Bokmål translation

jonassterud:
Here is a translation for Norwegian Bokmål.


leviport (approved):
Thanks for the translation!

jacobgkau:
@leviport I'm not seeing this work on a test machine with a user's language set to nb_NO:

Skjermdump fra 2021-08-10 08-26-17

When looking at the Portuguese translation that wasn't working, it seemed like some of the special characters may have needed to be contained within quotation marks; I'm wondering if this one has the same issue, or if I'm just doing something wrong testing. Did you test this and see it working?

jonassterud:
> You're right, I should have run it. I'm trying to request changes, but github is giving me a 500 error when I do so. Please don't merge this until that issue has been addressed.

I can add it if it isn't working for you. Quotes (") around every special character like this: bla"æ"bla?

jacobgkau:
> Quotes around every special character like this: bla"æ"bla?

Looking at the [Fluent documentation](https://projectfluent.org/fluent/guide/special.html), it looks like it would be `bla{"æ"}bla`, but testing locally, that doesn't seem to get it working. I'm also seeing special characters show up if I insert them into the English translation, so I'm not sure if that's actually the problem.

jacobgkau:
I tried setting the fallback language in `i18n.toml` to `nb_NO`, and found that while that language code was recognized, the translation library was looking for a directory called `nb-NO`. Moving the directory from `nb_NO` to `nb-NO` gets this working.

jacobgkau (approved):
Shows up in a user account set to Norwegian Bokmål, still shows up normally in an English user account as well.

jacobgkau:
The Portuguese translation was also not working, so I renamed that one to match, which should get it working.

jackpot51 (dismissed):

jackpot51 (approved):

jacobgkau (approved):

PR #207, opened 2021-08-10
fix(shlibs): Update version to 1

ids1024:
The soname version has changed since the version in `Cargo.toml` was changed to 1.0.0.

This fixes an error in `dh_shlibdeps` on my Gnome 40 branch of gnome-control-center.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ endif
 
 VENDORED ?= 0
 ifeq ($(VENDORED),1)
-	ARGS += "--frozen"
+	ARGS += "--frozen" "--offline"
 endif
 
 BINARY=target/$(TARGET)/$(BIN)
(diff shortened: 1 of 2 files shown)


jackpot51 (dismissed):

ids1024:
Oh joy, and the bionic and hirsute builds are failing with `attempting to update a git repository, but --frozen was specified`...

Makefile
```diff
@@ -25,7 +25,7 @@ endif
 
 VENDORED ?= 0
 ifeq ($(VENDORED),1)
-	ARGS += "--frozen"
+	ARGS += "--frozen" "--offline"
```
jackpot51, line 28: Only --offline should be required

jackpot51 (approved):

Makefile
jackpot51: That said, it is probably more safe to just pass both

ids1024: Interestingly, `--offline` doesn't seem to imply `--frozen`. In some cases, `--offline` builds can work without a `Cargo.lock` file already existing (when there's a local copy of the registry and dependent crates, and there are no git dependencies).

So perhaps it makes sense to pass both.

PR #211, opened 2021-09-01
Impish Release

mmstick:


jackpot51 (dismissed):

jacobgkau:
Attempting to upgrade a test machine from 21.04 with `pop-upgrade release upgrade -f`, I get:

```
Current Release: 21.04
Upgrading to: 21.10
New version available: false
Event: updating package lists
Release upgrade status: release upgrade aborted: failed to update package lists for the current release
pop-upgrade: calling ReleaseUpgradeFinalize method failed
  caused by: release upgrade cannot be finalized, because a release upgrade was not performed
```

Journalctl output: [journalctl.txt](https://github.com/pop-os/upgrade/files/7164573/journalctl.txt)

andrewda:
Getting the same error as @jacobgkau:

```
$ pop-upgrade release upgrade -f
Current Release: 21.04
Upgrading to: 21.10
New version available: false
Event: updating package lists
Event: fetching updated packages for the current release
Event: upgrading packages for the current release
Event: ensuring that system-critical packages are installed
Event: updating the source lists
Event: waiting on a process holding the apt lock files
Event: updating package lists
Event: updating package lists
Event: fetching packages for the new release
Event: simulating upgrade
Release upgrade status: release upgrade aborted: failure to simulate upgrade
pop-upgrade: calling ReleaseUpgradeFinalize method failed
  caused by: release upgrade cannot be finalized, because a release upgrade was not performed
```

jackpot51:
Getting the output of `journalctl -b 0 -t pop-upgrade` would help diagnose the issue

jackpot51:
@jacobgkau it would be good to try again, with only the `impish` staging repo added.

mmstick:
The logs show that `apt-cache rdepends snapd` returned an exit status of 100. But that's probably because of a source repo error.

jackpot51 (approved):

jacobgkau:
> Merging [#211](https://codecov.io/gh/pop-os/upgrade/pull/211?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) ([fe99b26](https://github.com/pop-os/upgrade/commit/fe99b26041d4bf92b3617526800b12a2e32888d6)) into [master](https://codecov.io/gh/pop-os/upgrade/commit/381fb2bb2b5500789095e40fd840b1dc29349ffd?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=None) ([381fb2b](https://github.com/pop-os/upgrade/commit/381fb2bb2b5500789095e40fd840b1dc29349ffd)) will **decrease** coverage by `0.89%`.

Are we/should we be paying attention to these Codecov comments at all? I tried to view the report to see what it's tracking, but I got an unauthorized/rate limit error.

image

If we're ignoring it, that's fine, just checking because we did decide to merge it in a while back.

mmstick:
It has no use here, honestly.

jacobgkau:
I'm still getting `failed to get transitional snap packages` on version fe99b26, fresh image of 21.04 trying to upgrade to 21.10.

jmbuhr:
I tried it as well. Does it maybe have something to do with it trying to reinstall / remove packages that it is not allowed to:
`E: Essential packages were removed and -y was used without --allow-remove-essential.`?

```
Sep 26 00:03:44 pop-os pop-upgrade[129264]: WARNING: The following essential packages will be removed.
Sep 26 00:03:44 pop-os pop-upgrade[129264]: This should NOT be done unless you know exactly what you are doing!
Sep 26 00:03:44 pop-os pop-upgrade[129264]:   pop-desktop pop-session (due to pop-desktop)
Sep 26 00:03:44 pop-os pop-upgrade[129264]:   pop-default-settings (due to pop-desktop)
Sep 26 00:03:44 pop-os pop-upgrade[129264]: 1359 upgraded, 62 newly installed, 9 to remove and 1 not upgraded.
Sep 26 00:03:45 pop-os pop-upgrade[129264]: E: Essential packages were removed and -y was used without --allow-remove-essential.
Sep 26 00:03:45 pop-os pop-upgrade[94802]: [ERROR] pop_upgrade src/release/mod.rs:616: failed to fetch packages: failure to simulate upgrade: status is unknown: exit code: >
Sep 26 00:03:45 pop-os pop-upgrade[94802]: [WARN ] pop_upgrade src/release/mod.rs:617: attempting to roll back apt release files
```

jacobgkau:
On 04e66ff, still failing with new output:
```
system76@pop-os:~$ pop-upgrade release upgrade -f
Current Release: 21.04
Upgrading to: 21.10
New version available: false
Event: updating package lists
Event: fetching updated packages for the current release
Event: upgrading packages for the current release
Event: ensuring that system-critical packages are installed
Event: updating the source lists
Event: waiting on a process holding the apt lock files
Event: updating package lists
Release upgrade status: systems are go for launch: reboot now
pop-upgrade: calling ReleaseUpgradeFinalize method failed
  caused by: release upgrade cannot be finalized, because a release upgrade was not performed
```

journalctl output: [out.txt](https://github.com/pop-os/upgrade/files/7245844/out.txt)

Strangely, the journalctl output continued accumulating after the CLI exited. It looks like packages continued being downloaded for 21.10 after the CLI's error message. Seems like the CLI and daemon aren't communicating properly.

jacobgkau:
aafe944 failed to build on the build server, but builds fine locally (but self-updates back to 381fb2b when running.)

jackpot51 (approved):

jacobgkau:
Upgrades appear to be working now, but upon finishing, I see a hang at `Upgrade complete. Now rebooting` (observed on pang11 & oryp8.) The system does not reboot by itself.

If I force power off and back on, I am running 21.10.

jacobgkau:
Not sure how relevant this is to the failure when doing the offline upgrade, but upon booting back into the partially-upgraded OS, this is the output I get from apt on both systems: [out.txt](https://github.com/pop-os/upgrade/files/7262043/out.txt)

It looks like some kind of network-manager/ppp dependency issue is causing the problem.

mmstick:
Everyone had the same issue when `ppp` was updated. I remember that we fixed it by doing an apt install fix with the  `--force-all` dpkg option. Latest commit now applies this.

data/pop-upgrade.sh
```diff
@@ -57,10 +57,10 @@ dpkg_configure () {
     # Validate the exit status
     dpkg --configure -a
 }
-
 apt_install_fix () {
     message -i "Checking for package fixes"
     env LANG=C apt-get -o Dpkg::Options::="--force-overwrite" \
+        -o Dpkg::Options::="--force-all" \
```
jackpot51, line 63: If you use force-all, you probably don't need the other two force options

mmstick: May as well apply it to all other install commands, too.

jackpot51: I don't think forcing all the dpkg force options is safe, though. If it is possible to find out which force option specifically fixes ppp upgrade then that would be better

mmstick: Reviewing the options, it must be one of these:

```
force-breaks
force-conflicts
force-depends
force-depends-version
```

mmstick:
Upgrades are successful now, but the system.sources file gets replaced with one that contains only the Ubuntu sources. Investigating

jackpot51:
@mmstick I would prefer the pop sources in a separate file. That should make what you describe not happen.

jackpot51 (approved):

jacobgkau:
Upgrading from a fresh 21.04 system seems to be working now, and the correct sources lists are in place after the upgrade. Once on 21.10, the Pop!_Shop fails to fetch updates with `Cannot refresh cache whilst offline`, which doesn't sound related to the package manager. However, while `sudo apt update` shows no updates available, `sudo apt upgrade` offers to downgrade some packages:
```
system76@pop-os:~$ sudo apt upgrade
Reading package lists... Done
Building dependency tree... Done
Calculating upgrade... Done
The following package was automatically installed and is no longer required:
  power-profiles-daemon
Use 'sudo apt autoremove' to remove it.
  libfirmware-manager libpop-upgrade-gtk libs76-hidpi-widget
The following packages will be DOWNGRADED:
  gir1.2-mutter-8 gnome-control-center gnome-control-center-data gnome-control-center-faces gnome-disk-utility
  gnome-shell gnome-shell-common gnome-shell-extension-prefs libdrm-amdgpu1 libdrm-common libdrm-intel1
  libdrm-nouveau2 libdrm-radeon1 libdrm2 libegl-mesa0 libgbm1 libgl1-mesa-dri libglapi-mesa libglx-mesa0
  libmutter-8-0 libxatracker2 linux-firmware mesa-va-drivers mesa-vdpau-drivers mesa-vulkan-drivers
  mutter-common
0 upgraded, 3 newly installed, 26 downgraded, 0 to remove and 0 not upgraded.
Need to get 235 MB/237 MB of archives.
After this operation, 20.1 MB of additional disk space will be used.
```
Any idea why this wasn't taken care of during `pop-upgrade`? (I'm doing some other tests with additional packages installed, so I'll see if it was just bad timing with packages being released.)

jacobgkau:
A second test upgrade succeeded:

- [X] 21.04 fresh install -> install all updates + this branch -> `pop-upgrade release upgrade -f` works for upgrading to 21.10.
- [X] 21.04 fresh install -> install wine/steam/kdenlive/obs-studio -> install this branch -> `pop-upgrade release upgrade -f` works for upgrading to 21.10.
  - [X] Self-updating to this version works.
  - [X] Attempting to upgrade without `-f` doesn't find a new release.
  - [X] Other updates are installed before downloading the release upgrade.

Still seeing those packages available to downgrade after the upgrade is complete, seems like something that needs to be fixed. `apt policy` for one of the packages says this:

```
system76@pop-os:~$ apt policy gnome-control-center
gnome-control-center:
  Installed: 1:40.0-1ubuntu5
  Candidate: 1:40.0-1ubuntu4pop0~1632861478~21.10~7ef4c24
  Version table:
 *** 1:40.0-1ubuntu5 500
        500 http://us.archive.ubuntu.com/ubuntu impish/main amd64 Packages
        100 /var/lib/dpkg/status
     1:40.0-1ubuntu4pop0~1632861478~21.10~7ef4c24 1001
       1001 http://apt.pop-os.org/release impish/main amd64 Packages
```

So it looks like the Ubuntu version's being installed during the upgrade somehow.

jackpot51:
It looks like the apt preferences (which have changed in pop-default-settings for 21.10 and later) are not in place during the upgrade.

leviport (approved):
From our testing, this is looking good now. We've done 18.04 upgrades all the way up through 21.10, tested Nvidia and integrated systems, and done upgrades with many different packages installed, and we aren't seeing issues.

PR #222, opened 2021-10-15
fix(upgrade.sh): After upgrade, set default session for each user to pop

mmstick:
This should resolve https://github.com/pop-os/beta/issues/236, in the event that an upgraded package sets itself as the default session.

### DIFF
--- a/data/pop-upgrade.sh
+++ b/data/pop-upgrade.sh
@@ -3,6 +3,13 @@
 export LANG=C
 export DEBIAN_FRONTEND="noninteractive"
 
+# Sets default session for each user as as pop
+set_default_session () {
+    for $file in /var/lib/AccountsService/users/*; do
+        sed "s/^Session=.*/Session=pop/g" -i $file
+    done
+}
+
 # Prevent apt sources from being reverted once this script launches
 rm -rf /pop-upgrade /pop_preparing_release_upgrade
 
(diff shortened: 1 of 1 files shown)


jacobgkau:
This doesn't seem to be working in its current state. With 774411e, I still got logged into LXQt after installing `lxqt-session` and doing an upgrade. When I checked `/var/lib/AccountsService/users/system76` for my user `system76`, the `Session=` line was empty (nothing after the equals sign.)

jacobgkau:
Okay, this issue is not specific to upgrades. If I do a clean install of 21.10, install lxqt-session, and reboot, my next login brings me to LXQt.

While I'm still confused by GDM not respecting the AccountsService files, I think I've tracked the Pop regression down to the gdm3 package. We have a patch applied in gdm3 to set `pop` as the fallback session: https://github.com/pop-os/gdm3/blob/master_impish/debian/patches/pop_prefer_pop_session_fallback.patch 

But my Impish machine's getting `gdm3` from Ubuntu instead of Pop:

```
system76@pop-os:~$ apt policy gdm3
gdm3:
  Installed: 41~rc-0ubuntu2
  Candidate: 41~rc-0ubuntu2
  Version table:
 *** 41~rc-0ubuntu2 500
        500 http://us.archive.ubuntu.com/ubuntu impish/main amd64 Packages
        100 /var/lib/dpkg/status
```

My Hirsute machine is getting `gdm3` from our PPA. So I think we just need to figure out why that package isn't in our PPA for Impish.

PR #227, opened 2021-12-13
fix(21.10): Remove /var/lib/bluetooth to resolve bluez issue

mmstick:
The journald logs for the pop-upgrade service should print a message about removing this directory after an upgrade is ready to be performed. The directory should be missing on shutdown.

### DIFF
--- a/src/release/mod.rs
+++ b/src/release/mod.rs
@@ -471,6 +471,11 @@ pub async fn upgrade<'a>(
         )
     }
 
+    if to == "21.10" {
+        info!("removing /var/lib/bluetooth to resolve https://github.com/bluez/bluez/issues/157");
+        let _ = fs::remove_dir_all("/var/lib/bluetooth");
+    }
+
     (*logger)(UpgradeEvent::Success);
     Ok(())
 }


jackpot51 (approved):

jacobgkau:
I was unable to confirm the issue described in https://github.com/bluez/bluez/issues/157 when upgrading from 21.04 to 21.10. I installed all 21.04 updates and rebooted before connecting a T3 speaker (along with an HP mouse) via Bluetooth, then upgraded to 21.10 without this fix. The T3 speaker still defaults to A2DP and is switchable between A2DP, HFP, and HSP on 21.10.

I also tried imaging 20.04, not installing any updates, connecting a pair of Anker headphones (with no Bluetooth mouse), and upgrading to 21.04 and then 21.10 without this fix. This device only gives me A2DP and HFP options on 21.10, but it still defaulted to A2DP and had no trouble switching between the two.

This is finally built, so I'll take it for a spin for regression testing now.

jacobgkau:
This branch doesn't release 21.10, so I'm having to test with `pop-upgrade release upgrade -f` to check that the directory is being removed.

On an upgrade from 21.04 to 21.10, I did not see the `/var/lib/bluetooth/` directory get removed. After the terminal shows `systems are go for launch: reboot now`, the permissions of the directory changed (it used to be readable by my user, but now I need `sudo` to read it), but it's still there along with its contents, although the modification time changed. Either the remove function is not working correctly, or something is recreating it almost immediately.

jacobgkau:
Since this is not working anyway, I wanted to share the concerns I was _planning_ to look for, which is 1) Bluetooth devices ceasing to work before I reboot for the upgrade, and 2) the directory not being restored if I cancel the upgrade after downloading (which should be a net no-op.)

Especially if something is recreating the directory, it seems like doing this in the offline `pop-upgrade.sh` script might be a safer bet. It seems like it should be able to check `/etc/os-release` after the package upgrades have been performed to decide whether or not to remove the directory.

(I was also wondering if we might want to add a warning or notification somewhere in the pop-upgrade-gtk copy to let the user know that Bluetooth devices will need to be re-paired after the upgrade, possibly only if the directory exists before the upgrade.)

jacobgkau:
Testing on another machine, I was able to remove the directory and it did not come back by itself, so it seems the remove function might not be working the way it's currently being used.

I opened a docs PR to document the fix command in the meantime: https://github.com/system76/docs/pull/864

PR #228, opened 2021-12-13
chore(changelog): Add changelogs for 21.10

mmstick:


changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+** Allow floating window exceptions to be tiled when toggled.
+** Transitions removed on opening a window to fix windows being stretched sometimes.
+** Dialogs will no longer unmaximize a maximized window when they gain focus.
+** Improved high contrast support.
+** Many improvements to the tiling behaviors of many unique window types.
+** Improvements to window attachment behaviors when new windows are spawned.
+
+* Pop! Shop
+** Many PRs have been submitted to elementary OS to address a number of problematic areas we've experienced.
+** The shop is currently in the process of being rebased, and will be released soon.
```
jacobgkau, line 30: This section doesn't seem like something that should be here. It's more of a roadmap item than a changelog item.

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+** Issues with packagekit and libflatpak causing freezes will be worked on in this release.
+
+* Pop! PPA now hosted by System76
+** A new packaging CI was developed in Rust for speedier package builds.
+** Which enables us to test fixes and improvements faster.
```
jacobgkau, line 35: This isn't a complete sentence, it's a dependent clause. It probably also shouldn't refer to "us" in the first person since the person reading it is not part of this group.

The four bullet points in this section can probably be combined into two:

> ** A new packaging CI was developed in Rust for speedier package builds, enabling faster testing of fixes and improvements.
> ** Packages are now published and delivered from apt.pop-os.org instead of Launchpad, allowing faster and more finely controlled delivery of updates.

I might also call this the "Pop!_OS repository" instead of "Pop! PPA" since PPAs are a Launchpad-specific term (afaik), but that one's not as important.

changelogs/21.10
jackpot51: Agree, only things that are done should be included

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+* Mainline Linux kernel
+** Starting with Linux 5.15, Pop!_OS will be shipping with the mainline kernel with the latest compatible NVIDIA drivers.
+** New kernel policy whereby the latest kernels will be released once they've passed extensive quality assurance tests.
+** System76 hardware lab contains a broad spectrum of chipsets, processors, and components for regression testing.
+
+* Improved System Stability When Installing & Upgrading Software
+** Apt is now prevented from enabling the accidental removal of critical desktop packages.
+** Idea pitched to Debian and Debian issued a similar fix that we now ship.
+
+* Quality Assurance
```
jacobgkau, line 48: As much as we appreciate the shout-out, this section doesn't really contain any concrete info, and probably shouldn't be referenced here in relation to 21.10 specifically.

isantop:
We appear to be missing information about the new GTK4 support in the theme for system applications.

mmstick:
@isantop Could you write what I should add here?

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+** Emoji picker set back to Ctrl + Shift + E
++* Alt+F4 added as an alternative to Super+Q for closing windows
+
+* Application Library
+** Replaces the GNOME application overview, achieving the same information density with less space.
+** Improves multi-monitor experience by launching onto the display that has mouse focus.
+** Alphabetical application sorting will make it easier to locate the application or folder you want.
+** Drag and drop applications into custom folders which are visible and act as tabs.
+** Search quickly filters installed applications and applications that are available to install from the Pop!_Shop.
+** Achieves the same information density while using much less space.
```
jacobgkau, line 16: This line is repeating the first line in this section.

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+** Various scenarios causing crashes were resolved.
+** Allow floating window exceptions to be tiled when toggled.
+** Transitions removed on opening a window to fix windows being stretched sometimes.
+** Dialogs will no longer unmaximize a maximized window when they gain focus.
+** Improved high contrast support.
+** Many improvements to the tiling behaviors of many unique window types.
+** Improvements to window attachment behaviors when new windows are spawned.
+
+* Pop! Shop
+** Many PRs have been submitted to elementary OS to address a number of problematic areas we've experienced.
```
jacobgkau, line 29: May want to reference "upstream" so someone not familiar with where the Pop!_Shop comes from will know why we're talking about elementary.

> ** Submitted several PRs upstream to the elementary OS project addressing problematic areas.

(If we could get more specifics than "problematic areas" from @isantop or anyone else, that would be good too.)

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+** Which allows us to deliver updates quicker, with more control over package builds.
+
+* Mainline Linux kernel
+** Starting with Linux 5.15, Pop!_OS will be shipping with the mainline kernel with the latest compatible NVIDIA drivers.
+** New kernel policy whereby the latest kernels will be released once they've passed extensive quality assurance tests.
+** System76 hardware lab contains a broad spectrum of chipsets, processors, and components for regression testing.
+
+* Improved System Stability When Installing & Upgrading Software
+** Apt is now prevented from enabling the accidental removal of critical desktop packages.
+** Idea pitched to Debian and Debian issued a similar fix that we now ship.
```
jacobgkau, line 46: Not sure if this line belongs here-- the previous line covers the actual change, this is just part of the story of how we got there.

changelogs/21.10
```diff
@@ -0,0 +1,72 @@
...
+
+* Quality Assurance
+** Quality assurance tests and capabilities have expanded.
+** Testing is now more extensive than in prior releases, and will continue to improve.
+
+* Upgrade Hardening
+** Various issues with relase upgrades were discovered and resolved.
+** User-added PPAs are disabled on upgrade because they are a often the cause of failed upgrades.
+** Fixed errors revolving around custom edits to the /etc/fstab file.
+** The recovery partition is now upgraded prior to updating the OS.
```
jacobgkau, line 56: I would either combine this with the next line like so:

> ** The recovery partition is now upgraded prior to upgrading the OS so it can be used to perform a refresh install in case of an upgrade failure.

Or add the word "now" to the next line to match:

> ** In the unlikely event of a failed upgrade, the refresh OS option will now install the new release instead of the old one.

changelogs/21.10
```diff
@@ -0,0 +1,61 @@
...
+** In the unlikely event of a failed upgrade, the refresh OS option will now install the new release instead of the old one.
+
+* Raspberry Pi Tech Preview
+** ISOs are now being generated for the Raspberry Pi 4 / 400.
```
jackpot51, line 49: SD card images, not ISOs

isantop:
@mmstick 

** The Pop GTK Theme has been updated with new support for system-wide GTK4 applications.
** The GNOME Shell theme has been updated to improve it's consistency and visual appearance.

jacobgkau:
Merged into the `impish` release branch so it can be tested (this branch would not show the upgrade or changelog in the GUI.)

PR #229, opened 2021-12-14
fix: Only upgrade recovery partition if it exists

jacobgkau:


leviport (approved):
- [x] 20.04 -> 21.10 in VM
- [x] 21.04 -> 21.10 in VM
- [x] 20.04 -> 21.10 on gaze10 in legacy mode
- [x] 21.04 -> 21.10 on gaze10 in legacy mode
- [x] 20.04 -> 21.10 on darp6
- [x] 21.04 -> 21.10 on darp6

Probably want to do a little extra testing on efi installs in the morning, but this is looking solid so far.

jackpot51 (approved):

PR #232, opened 2021-12-14
fix(restore): Remove impish sources when cancelling on hirsute or focal

### LINKED ISSUE #193: Pop!_OS PPA not rolled back after canceling upgrade
**Steps to recreate:**
1. On a Groovy system, download the upgrade.
2. After downloading has completed, click Upgrade.
3. In the changelog dialog, click Cancel.
4. After `Canceling upgrade` has gone away, `system76-ubuntu-pop-groovy.list` and `pop-os-ppa.list` are both present and have active lines. The first list contains a `groovy` entry and the second list contains a `hirsute` entry.
5. Run `sudo apt update` and observe that Hirsute packages are available for installation.

**Expected behavior:**
This is a little messy because the installer on the ISO and pop-upgrade use two different default filenames for this file. Some scenarios that might work:
- `pop-os-ppa.list` could be removed if `system76-ubuntu-pop-x.list` is restored.
- `pop-os-ppa.list` could be edited to the current distribution version when canceling an upgrade (which would at least prevent further issues), and `system76-ubuntu-pop-x.list` could then optionally be left out from the backup/restore.

mmstick:
If there are `.save` files to restore from.

Closes #193

### DIFF
--- a/src/release/mod.rs
+++ b/src/release/mod.rs
@@ -374,18 +374,18 @@ pub async fn upgrade<'a>(
 
     let _ = AptMark::new().hold(&["pop-upgrade"]).await;
 
+    let version = codename_from_version(from);
+
     // Check the system and perform any repairs necessary for success.
     (async move {
         repair::crypttab::repair().map_err(RepairError::Crypttab)?;
         repair::fstab::repair().map_err(RepairError::Fstab)?;
-        repair::packaging::repair().await.map_err(RepairError::Packaging)?;
+        repair::packaging::repair(version).await.map_err(RepairError::Packaging)?;
 
         Ok(())
     })
     .await
     .map_err(ReleaseError::Repair)?;
-
-    let version = codename_from_version(from);
 
     info!("creating backup of source lists");
     repos::backup(version).map_err(ReleaseError::BackupPPAs)?;
--- a/src/release/repos.rs
+++ b/src/release/repos.rs
@@ -85,7 +85,7 @@ pub fn backup(release: &str) -> anyhow::Result<()> {
     Ok(())
 }
 
-fn delete_system76_ubuntu_ppa_list() -> anyhow::Result<()> {
+fn delete_system76_ubuntu_ppa_list() {
     if let Ok(ppa_directory) = Path::new(PPA_DIR).read_dir() {
         for entry in ppa_directory.filter_map(Result::ok) {
             let path = entry.path();
@@ -93,20 +93,17 @@ fn delete_system76_ubuntu_ppa_list() -> anyhow::Result<()> {
                 if let Some(fname) = path.file_name() {
                     const POP_PPA: &[u8] = b"system76-ubuntu-pop";
                     if fname.to_raw_bytes().windows(POP_PPA.len()).any(|w| w == POP_PPA) {
-                        fs::remove_file(&path).context("failed to remove the old Pop PPA file")?;
-                        return Ok(());
+                        let _ = fs::remove_file(&path);
                     }
                 }
             }
         }
     }
-
-    Ok(())
 }
 
 /// For each `.list` in `sources.list.d`, add `#` to the `deb` lines.
 pub fn disable_third_parties(release: &str) -> anyhow::Result<()> {
-    delete_system76_ubuntu_ppa_list()?;
+    delete_system76_ubuntu_ppa_list();
     let dir = fs::read_dir(PPA_DIR).context("cannot read PPA directory")?;
     for entry in iter_files(dir) {
         let path = entry.path();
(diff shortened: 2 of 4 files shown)


jackpot51 (approved):

jacobgkau (changes requested):
This doesn't seem to be working. After adding this branch on 21.04, downloading the upgrade, and clicking `Upgrade` and then `Cancel`, I still have impish apt.pop-os.org mixed in with my restored hirsute sources:
```
system76@pop-os:~$ ls -alh /etc/apt/sources.list.d/
total 28K
drwxr-xr-x 2 root root 4.0K Dec 15 10:55 .
drwxr-xr-x 7 root root 4.0K Dec 15 10:55 ..
-rw-r--r-- 1 root root  197 Dec 15 10:53 popdev-restore.list
-rw-r--r-- 1 root root  125 Dec 15 10:55 pop-os-apps.sources
-rw-r--r-- 1 root root  202 Dec 15 10:55 pop-os-ppa.list
-rw-r--r-- 1 root root  139 Dec 15 10:53 pop-os-ppa.sources
-rw-r--r-- 1 root root  291 Dec 15 10:55 system.sources
system76@pop-os:~$ sudo apt update
[sudo] password for system76: 
Hit:1 http://apt.pop-os.org/proprietary hirsute InRelease
Hit:2 http://apt.pop-os.org/release impish InRelease                     
Hit:3 http://us.archive.ubuntu.com/ubuntu hirsute InRelease              
Hit:4 http://apt.pop-os.org/staging/restore hirsute InRelease            
Hit:5 http://us.archive.ubuntu.com/ubuntu hirsute-security InRelease     
Hit:6 http://us.archive.ubuntu.com/ubuntu hirsute-updates InRelease      
Hit:7 http://ppa.launchpad.net/system76/pop/ubuntu hirsute InRelease
Hit:8 http://us.archive.ubuntu.com/ubuntu hirsute-backports InRelease
Reading package lists... Done
Building dependency tree... Done
3 packages can be upgraded. Run 'apt list --upgradable' to see them.
```

mmstick:
Added logs and an extra conditional that's more explicit about removing this file.

jacobgkau:
On 3dac2c6, this is behaving as expected from a fresh 21.04 installation when cancelling a 21.10 installation. However, the logs do not include any "removing" lines. The logs only  include "restoring source list as" lines.

To make sure I'm not misunderstanding what lines 186-197 are trying to do, I downloaded the upgrade, touched a `fake.list` file (without a corresponding `.save` file), and then cancelled the upgrade. The `fake.list` file was not removed. (This would be what was happening when the `pop-os-ppa.sources` file wasn't removed before its explicit removal was added on lines 243-245.)

I should be seeing lines for the non-.save files in the logs being removed, correct?

mmstick:
You should, yes. Confusing that it's not even detecting that there are save files. I created a unit test and it checks out that paths that have the save extension are correctly identified.

jacobgkau:
I don't see the `Removing sources which lack .save backups` line on 3dac2c6, either, if that helps to know.

mmstick:
Rebased with only the commit that explicitly removes the impish sources on hirsute / focal restores.

jacobgkau:
Can we not just get the removal working correctly so that workaround isn't needed? I've been waiting for e52a373 to build, it looks like the build server is getting stuck again today.

mmstick:
The other commits have been moved to the `save` branch. Whichever gets  fix out sooner.

jacobgkau:
8b9284f fixes the current issue with rollback on 21.04, but it doesn't fix the issue for 20.04 LTS. (On 20.04 LTS, after downloading and then cancelling, I have `focal` Ubuntu/Pop PPA/Pop Apps in the main `sources.list`, `impish` Ubuntu/Pop Apps in their respective `.sources` files, and a duplicate `focal` Pop PPA in its own `system76-ubuntu-pop-focal.list` file.)

jacobgkau:
This is almost working on 20.04 now, but I still have duplicate sources after the rollback that were not there before. The duplicate is between `/etc/apt/sources.list` and `/etc/apt/sources.list.d/system76-ubuntu-pop-focal.list`, which seems to be the PPA.

jacobgkau:
The duplicate sources after cancelling/rolling back on Focal are now fixed.

This also seems to fix https://github.com/pop-os/upgrade/issues/238 now. Cancelling/rolling back does not add PPAs back when it theoretically should (the expected behavior is for the state before upgrade and after cancel to be identical), but this is still a net improvement. Still testing for any other issues.

jacobgkau (approved):
I believe the discrepancy with restoring PPAs has to do with where they're stored. Because Focal uses `sources.list` for its default repos, any custom sources in `sources.list` are wiped out when the defaults are applied after a rollback. I'm not considering this blocking right now, it's a much more trivial problem than the issues currently in the wild.

This resolves high-priority issues https://github.com/pop-os/upgrade/issues/193 and https://github.com/pop-os/upgrade/issues/238, which should reduce the number of users experiencing libc6/ppp/other dependency issues.

Performed some test upgrades from 21.04, 20.04, and 18.04; there do not seem to be any regressions handling a variety of third-party and non-default packages.

jackpot51 (approved):

Issue #243, opened 2021-12-29
Can't upgrade to latest release from 20.10


SF97:
Hi

I'm currently on pop os 20.10 and trying to update to latest version by running:

`pop-upgrade release upgrade`

However the following errors occurs:

![image](https://user-images.githubusercontent.com/10375710/147617559-cf466a76-1400-49ef-adad-eba1d412b5e3.png)

I checked pop-upgrade logs with journalctl:

`journalctl -b0 -o cat -u pop-upgrade`

and the following error is shown:

```
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libpop-upgrade-gtk : Depends: libc6 (>= 2.33) but 2.32-0ubuntu3 is to be installed
                      Depends: libgdk-pixbuf-2.0-0 (>= 2.22.0) but it is not installable
 pop-upgrade : Depends: libc6 (>= 2.33) but 2.32-0ubuntu3 is to be installed
E: Unable to correct problems, you have held broken packages.
pop-upgrade.service: Main process exited, code=exited, status=1/FAILURE
pop-upgrade.service: Failed with result 'exit-code'.
pop-upgrade.service: Scheduled restart job, restart counter is at 1.
Stopped Pop Upgrade Daemon.
Started Pop Upgrade Daemon.
```

It looks like libc6 version >= 2.33 is needed, however that version is not available for my pop os version (20.04). For that reason libpop-upgrade-gtk can not be updated to the latest version and the upgrade process fails

If I understand correctly, one of the first steps of the upgrade is making sure that pop-upgrade is on the latest version and this is impossible on the version I have since the dependencies are not available

How can I fix this?

jacobgkau:
Please provide the output of the following commands:

```
cat /etc/os-release
sudo apt update
apt policy pop-upgrade
```

This is a strange issue to be having now, since the last release of pop-upgrade for 20.10 was made in July of last year (no further updates can be published now, because Launchpad does not allow publishing for end-of-life Ubuntu versions.) If some of your apt sources are not pointing to the correct 20.10 release (which the above commands should show, and which may have been caused by a pop-upgrade issue after a prior upgrade attempt), that could explain why this is happening.

[...]

Issue #164, opened 2021-05-24
On BIOS machine, after upgrading from 18.04 to 20.04, subsequent upgrade to 20.10 hangs


jacobgkau:
Steps to recreate:
- Perform clean installation of Pop!_OS 18.04 LTS.
- Install all available updates via Pop!_Shop.
- Upgrade to 20.04 LTS.
- Attempt to upgrade to 20.10.

Actual behavior:
- After clicking `Download`, the interface hangs at `Downloading Pop!_OS 20.10`. Checking journalctl shows that `grub-pc` is asking an interactive question and pop-upgrade is not answering:

![Screenshot from 2021-05-20 11-43-44](https://user-images.githubusercontent.com/7199422/119033938-0dda9480-b96b-11eb-9667-f20cad21898e.png)

This is occurring on the `master` and `rec` branches.

jacobgkau:
I ran into this again today while testing https://github.com/pop-os/upgrade/pull/244 and running `pop-upgrade release upgrade`. It seemed to happen when pop-upgrade was self-updating itself (it started with what's currently released on 20.04, and needed to be upgraded to the staging branch's version.)

If I install the pop-upgrade update manually instead of letting it self-update, I get a TUI prompt for GRUB:

![image](https://user-images.githubusercontent.com/7199422/148297051-5d8c7580-db32-4d5d-9a75-e4bd2c7870bf.png)

After finishing that prompt, I was able to run pop-upgrade successfully and it didn't hang.

PR #244, opened 2021-12-31
Upgrade Improvements

mmstick:
- Show progress for prerequisites being installed during the offline upgrade, to prevent people from assuming that nothing is happening when this process takes a lot of time to complete.
- Be explicit about declaring the candidate version of prerequisites to install when doing that prerequisite upgrade, to ensure that apt knows that we really do want the latest candidate installed.
- Cleans up the way prerequisites are installed, with all packages determined required to be appended to an array for a single install invocation, and adds ppp and libm0 to the list of packages that need the extra handholding at this stage.
- Adds similar repair fix for the `pop-upgrade repair` routine, which ensures that candidate versions of each of these packages are installed when conflicts occur.
- Improvements to old-releases handling when repairing, upgrading, and restoring.


jacobgkau:
Tested an 18.04 -> 20.04 upgrade in a VM using the problem package list from [this comment](https://github.com/pop-os/upgrade/pull/142#issuecomment-832205986), then 20.04 -> 21.10 on the same installation. https://github.com/pop-os/upgrade/issues/164 is still occurring, but that's not a regression.

Even after I worked around #164, the 20.04 -> 21.10 upgrade failed. The system did not power off/reboot properly after the offline upgrade, and upon booting again, I had Focal packages with Impish sources. I will attempt to replicate this from a fresh 20.04 installation using a modified version of the package list.

jacobgkau:
I recreated the failure from a fresh 20.04 LTS installation (with all updates applied) using this package list: [installed-focal-oneline.txt](https://github.com/pop-os/upgrade/files/7823847/installed-focal-oneline.txt)

Here is what I see after `Upgrade complete. Now rebooting`:

image

(It seems this output is normal in a VirtualBox installation, but the system does not finish powering off/rebooting.)

Upon rebooting, GDM is not using the Pop theme, and I have mixed package versions (e.g. `pop-default-settings` is on a 20.04 version, but `gdm3` is on a 21.10 version.)

Currently performing the upgrade one more time on stable pop-upgrade to confirm this is a regression.

jacobgkau:
Hmm... I thought this worked the very first time I tried on stable, but I seem to have encountered the same issue testing on stable now. I'll try running the upgrade script manually to see where it might be having an issue.

jacobgkau:
Here are logs from running `pop-upgrade release upgrade` and then `sudo /usr/lib/pop-upgrade/upgrade.sh` on this branch. The only change I made to `upgrade.sh` before running was removing the line to reboot.

[upgrade-1.txt](https://github.com/pop-os/upgrade/files/7824071/upgrade-1.txt)
[upgrade-2.txt](https://github.com/pop-os/upgrade/files/7824072/upgrade-2.txt)

jacobgkau:
6197b64 seems to fail faster. upgrade.sh output: [upgrade.txt](https://github.com/pop-os/upgrade/files/7849058/upgrade.txt)

mmstick:
With the last commit, pop-upgrade is no longer burdened by the cursed package list.

jacobgkau:
Confirmed the list from https://github.com/pop-os/upgrade/pull/142#issuecomment-832205986 is no longer killing the 20.04 -> 21.10 upgrade!

I noticed that if `chromium-browser` and `chromium-codecs-ffmpeg-extra` were installed on 20.04, after the upgrade to 21.10 is complete, `chromium-browser` immediately has an update available to Ubuntu's version (and installing that update pulls in the snap, while our outdated `chromium` package is also still installed.) Confirmed this isn't a regression.

jacobgkau:
While 66b2eb3 fixed the upgrade from 20.04 -> 21.10, it seems to have broken the upgrade from 18.04 -> 20.04, with the package list.

mmstick:
I'll see what I can do about that

jacobgkau:
I'm running the script manually now, I saw `Command line option 'y' [from '-y'] is not understood` near the beginning, in case that's related.

jacobgkau:
Here was the log from 66b2eb3: [upgrade-output-1804.txt](https://github.com/pop-os/upgrade/files/7857182/upgrade-output-1804.txt)

jacobgkau:
Still failing on 412ac24, here is the output: [upgrade-1804-output.txt](https://github.com/pop-os/upgrade/files/7858297/upgrade-1804-output.txt)

jacobgkau:
https://github.com/pop-os/upgrade/issues/164 is still happening as mentioned earlier (still not a regression), but 6dbf517 works for 18.04 -> 20.04 and 20.04 -> 21.10 in a virtual machine.

jacobgkau:
Just to be clear, I'm not sure if I can actually test whether the attempted fixes for https://github.com/pop-os/upgrade/issues/164 are working, since I'm on stable pop-upgrade after the 18.04 -> 20.04 upgrade and it's the old version that tries to install the new version from there. Unless we know of a way to artificially induce the `grub-pc package is being upgraded` screen, in which case I could test the self-upgrade from the branch to stable after removing the staging repo.

Still doing regression testing aside from that.

jacobgkau:
I guess the question might be why this prompt is showing up even though the `grub-pc` update should have happened during the offline portion of the upgrade.

mmstick:
debconf dialogs can show up even if you define `noninteractive` because `dpkg` doesn't listen to that setting. It does require setting `--force-confdef` to have it assume the default configuration in the file, and `--force-confold` to prefer existing configs if they already exist.

Unsure on the best way to induce the dialog. It's commonly seen each time it updates when there's no pre-existing grub config.

jacobgkau:
This doesn't seem to fix https://github.com/pop-os/upgrade/issues/243, is it supposed to? (The description mentions old-releases improvements.)

On a 20.10 virtual machine, I added the Focal staging branch (since this branch is not being built for Groovy), then tried to run pop-upgrade. Here's the output: [groovy.txt](https://github.com/pop-os/upgrade/files/7865144/groovy.txt)

Not sure if https://github.com/pop-os/upgrade/issues/116 is related to this since it looks like the configured staging branch was reverted from focal to groovy at some point in the process, but I'm also seeing strange behavior with the old-releases URLs sometimes working and sometimes not when just running `apt update` (possibly seems like it doesn't work for a minute after updating the repo config?): [groovy-apt-update.txt](https://github.com/pop-os/upgrade/files/7865145/groovy-apt-update.txt)

I'm inclined to move forward with this PR as-is to get the other fixes in, just keeping track of what is and isn't addressed.

mmstick:
Sure, we can have it move forward as-is. There's an attempt to fix old-releases. Will follow up with a PR in the future for that.

jacobgkau (approved):
I've tested d989c8b in the following configs, all successful:
- 18.04 -> 20.04 -> 21.10 with the extensive cursed package list in a VM
- 20.04 -> 21.10 with wine, kdenlive, steam, and lutris on a physical machine
- 21.04 -> 21.10 with wine, kdenlive, steam, lutris, ardour, and vlc on a physical machine

Contains some good fixes and no regressions from what's currently published.

PR #245, opened 2022-01-01
add clippy msrv config

danieleades:
adding a clippy 'msrv' config prevents clippy from suggesting changes that use syntax not supported by the chosen compiler version.

this isn't strictly necessary since the compiler version is pinned by the `rust-toolchain` file, which also constrains the version of clippy that is run. What this allows though is running a later clippy version to check warnings from newer clippy lints, but that are still compatible with the chosen version (`cargo +nightly clippy`)


mmstick:
We have control over Rust packaging in Pop so this isn't necessary anymore. I'll be keeping Rust to the latest stable version as needed. We currently have 1.56 and I'll likely upgrade it to 1.58 as soon as it releases.

PR #248, opened 2022-01-03
enforce formatting in CI

danieleades:
is this a reasonable opportunity to re-introduce this?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,24 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+
+name: Continuous integration
+
+jobs:
+
+  fmt:
+    name: format
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


mmstick:
Nah, we have our own CI

jacobgkau:
I'd like to point out that we've had to start adding skips to ignore this formatting check in other projects where it's been added: https://github.com/pop-os/system76-power/pull/290#discussion_r743994028

danieleades:
> I'd also like to point out that we've had to start adding skips to ignore this formatting check in other projects where it's been added (see [pop-os/system76-power#290 (comment)](https://github.com/pop-os/system76-power/pull/290#issuecomment-962225465) and the following resolved discussion.)

Method chaining and postfix comments don't always play nicely together. I suspect the comments are ignored when calculating the hueristics. It's generally more idiomatic to use prefix comments interspersed between the chained methods

PR #261, opened 2022-02-04
Resumable downloads & cancelability improvements

mmstick:
Utilizes our async-fetcher crate for fetching the recovery ISO. You can test it by running

```
pop-upgrade recovery upgrade from-release
```

Then `Ctrl + C` somewhere in-between the download to cancel the process.

Running the same command will resume where it left off.

Refactors some of the daemon service to utilize and support tokio. Long term dbus and dbus-crossroads should be replaced with zbus, and the service rearranged to be more async-friendly.


jackpot51 (approved):

jacobgkau:
1d85ded failed to build on all platforms. Also fails to build locally.

```
error[E0107]: this struct takes 0 lifetime arguments but 1 lifetime argument was supplied
   --> src/main.rs:193:26
    |
193 | async fn main_(matches: &ArgMatches<'_>) -> anyhow::Result<()> {
    |                          ^^^^^^^^^^---- help: remove these generics
    |                          |
    |                          expected 0 lifetime arguments
    |
note: struct defined here, with 0 lifetime parameters
   --> /home/system7/upgrade/vendor/clap/src/parse/matches/arg_matches.rs:71:12
    |
71  | pub struct ArgMatches {
    |            ^^^^^^^^^^

error[E0581]: return type references lifetime `'_`, which is not constrained by the fn input types
   --> src/main.rs:193:45
    |
193 | async fn main_(matches: &ArgMatches<'_>) -> anyhow::Result<()> {
    |                                             ^^^^^^^^^^^^^^^^^^

Some errors have detailed explanations: E0107, E0581.
For more information about an error, try `rustc --explain E0107`.
warning: `pop-upgrade` (bin "pop-upgrade") generated 20 warnings
error: could not compile `pop-upgrade` due to 2 previous errors; 20 warnings emitted
make[2]: *** [Makefile:75: target/release/pop-upgrade] Error 101
make[2]: Leaving directory '/home/system7/upgrade'
make[1]: *** [debian/rules:18: override_dh_auto_build] Error 2
make[1]: Leaving directory '/home/system7/upgrade'
make: *** [debian/rules:7: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
```

mmstick:
Should be fixed now.

jacobgkau:
If I start the upgrade from a terminal, cancel it with Ctrl-C, then open the GUI and try to download it there, I get a `process has been cancelled` error:

Screenshot from 2022-02-07 12-22-16

It seems to work the second time I try after having cancelled, but not the first. (Also noticed while testing that the CLI progress incorrectly goes to 100%/fully fetched after cancelling partway through, as is visible in the screenshot.)

jacobgkau:
On 5b0240f, both of the described issues are resolved. Resuming from cancelled downloads is still working.

While testing behavior when switching network connections (e.g. plugging in/unplugging Ethernet while also connected to WiFi), I saw that sometimes (after the ~second connection change and onwards) the reported progress is doubled when the download starts progressing again. In the GUI, this results in impossible progress of greater than 100% being reported:

Screenshot from 2022-02-07 13-30-55

In the CLI, the download seemed to stop when the incorrect progress number reached 100%, followed by errors in journalctl about the recovery files failing to extract:

Screenshot from 2022-02-07 13-34-26

Powering off and then attempting to access Recovery with the spacebar shows that it was rendered unbootable by that error.

After the errors (before powering off), I also had three of what seem to be the bad recovery image mounted in Nautilus:

Screenshot from 2022-02-07 13-35-10

mmstick:
Automatic retries aren't resetting progress. Will fix.

mmstick:
The progress should now get properly reset when a retry is being performed. The the result of the recovery fetch task wasn't being returned, which caused it to believe that the task was Ok.

jacobgkau:
On a103309 with the CLI, the number fetched is now going back down to 0 when I cancel with Ctrl-C. That's less of a misrepresentation than jumping to 100%, but still inaccurate and could be annoying if someone's trying to troubleshoot.

jacobgkau:
Progress reporting appears to be working correctly, or at least much better, now. The CLI keeps its current number when I cancel with Ctrl-C.

After canceling/resuming and switching network connections a few times, I eventually got a `failed to download ISO` error message after unplugging Ethernet:

Screenshot from 2022-02-08 09-56-33

I clicked Update again, and the download resumed from where it was (near the end), but once it finished, the checksum was invalid:

Screenshot from 2022-02-08 09-57-25

After that, when I click Update again, it seems to just try and verify what's already downloaded again (which fails again):

Screenshot from 2022-02-08 09-59-57

Screenshot from 2022-02-08 10-00-03

We might need to delete the iso if the checksum verification fails? (As for why it was corrupted in the first place, it _seemed_ like the progress might have been jumping slightly after some network change retries, but the amount was so small that it could have also just been e.g. the last bit of progress from before the disconnection not being reported until it's reconnected, so I can't say for certain.)

jacobgkau:
The first time I tried to update recovery on 0ce6eab, I got the "failed to verify" error again, but the second time, it started downloading from scratch, so it looks like it removes after failure to verify now. I guess it makes sense that we don't want to automatically retry indefinitely, so this behavior looks good.

When resuming after cancelling, I am sometimes seeing a large delay/hang when I try to resume the download:

Screenshot from 2022-02-08 11-07-45

When I tried to Ctrl-C from that the first time it happened, I continued seeing MessageCall log messages for a minute, although it did eventually exit:

Screenshot from 2022-02-08 11-08-19

On trying again, the upgrade resumed as I would expect; the next time I resumed, it hung again and didn't exit even after waiting many minutes after pressing Ctrl-C. Every time I pressed Ctrl-C, I saw another `cancelling a process which is in progress` log message, but it never actually exited.

The hang seems to always happen the first time I resume after switching network connection, e.g. to recreate:

1. Start the download.
2. Switch network connection (plug in/unplug Ethernet).
3. Wait for the download to resume and start progressing again.
4. Cancel manually with Ctrl-C.
5. Try to resume by running the command again.
6. Observe hang.

The one where it never exited was on WiFi, so I'm wondering if network connection/speed has something to do with it. The last one that I recreated on Ethernet exited immediately when I pressed Ctrl-C, with a different error message in the logs:

Screenshot from 2022-02-08 11-28-36

I had never seen a hang at this point until 0ce6eab, so it could be related to the version of async-fetcher that was changed with that commit.

mmstick:
The change in that commit is that it flushes the file when canceled. Perhaps I should have it flush more often, or do the flush in the background.

jacobgkau:
Still able to create a hang using the steps listed before on d225885.

mmstick:
Testing out a fix but I think it's working. Will update if it does.

jacobgkau:
The hang described in https://github.com/pop-os/upgrade/pull/261#issuecomment-1032933185 is no longer occurring! :tada: 

On a0c6cd2, I am now unable to cancel with Ctrl-C in the terminal when downloading over Ethernet (seeing the progress go up by about 100 MiB/s). After attempting to cancel, the progress continues going up until I unplug the Ethernet cable, at which point it finally fails/cancels with `recovery upgrade aborted: failed to download iso`. This does not happen on WiFi; when I press Ctrl-C on WiFi, the download stops with that message after just a few seconds, without having to disconnect.

On WiFi, the progress still goes up at least one more time after I start the cancel operation. I assume the issue on Ethernet has something to do with the much higher speed (whatever mechanism is supposed to cancel not having time to run/complete.)

mmstick:
I think you're right about the higher speed not giving time to yield to the next future. I'll add some yield calls and see if that helps.

mmstick:
I'm on a decent 16 MiB/s Ethernet connection right now and can cancel as it's downloading, and while the network is disconnected.

jacobgkau:
I am able to cancel on Ethernet again on 7437eec, but the freeze from https://github.com/pop-os/upgrade/pull/261#issuecomment-1032933185 is now happening again.

mmstick:
I'll ping you when I see all of these instances resolved.

mmstick:
Okay, I'm not seeing any hangs when canceling during network loss, when canceling after networking has been restored, or when resuming after either of those conditions.

jacobgkau:
> Okay, I'm not seeing any hangs when canceling during network loss, when canceling after networking has been restored, or when resuming after either of those conditions.

I know that wasn't a ping, but I'm still seeing the hang on fbdad21... after upgrading pop-upgrade, I rebooted, confirmed the pop-upgrade version, started download on WiFi, cancelled, resumed on WiFi, plugged in Ethernet, cancelled, and attempted to resume.

jacobgkau:
In case it's useful, when it gets into this "hung" state, if I try to cancel with Ctrl-C, it still waits a while, but it does seem to exit eventually-- and much quicker on Ethernet than WiFi. I get the `failed to find build` messages as shown in the last screenshot of https://github.com/pop-os/upgrade/pull/261#issuecomment-1032933185 when it finally does exit after cancelling from a hang.

mmstick:
That is a good clue

mmstick:
@jacobgkau Cancelations will now return errors and show them correctly in the client.

jacobgkau:
Looks better now:
Screenshot from 2022-02-10 11-42-18

CLI:
- [X] Able to cancel/resume with appropriate output.
- [X] Download resumes when changing network connection.
- [X] Able to cancel/resume again after changing network connection.
- [X] Download resumes when changing network connection in the other direction.
- [X] Still able to cancel/resume with appropriate output.

GUI:
- [X] Able to resume.
- [X] Able to change network connection in both directions.
- [X] Closing the window continues downloading in the background; progress is reported again on re-opening.

I now had two problems near the end of testing:
- On the last network reconnection, a timeout was reached for some reason (I'm not sure why, as I just unplugged Ethernet like I'd already done a number of times before without getting this error):
Screenshot from 2022-02-10 11-46-32
That one's not a blocking problem (it makes sense the download might stop/fail if the network cable is unplugged), just thought I'd mention it.

- After cancel/resuming and reconnecting a number of times, I again had a checksum mismatch:
Screenshot from 2022-02-10 11-47-05
This one seems like it could be indicative of an issue with the resume capability. Is there anywhere resuming might not be picking back up/writing correctly?

mmstick:
This change may help. Canceling the fetcher shouldn't abruptly cancel writes to the ISO that are mid-write. The task will complete the write that it is currently performing, and flush the file before exiting.

jacobgkau:
Still got a corrupted iso with the same test procedure. Interestingly, I got a timeout this time just before the download finished, even though I hadn't changed network connection immediately before the timeout:
Screenshot from 2022-02-10 13-05-07

When I clicked Update again, it almost immediately started verifying the image, but the checksum didn't match:
Screenshot from 2022-02-10 13-05-19

mmstick:
Maybe there's a bug in the HTTP client I'm using with its own internal timeouts. I'll try disabling the `low_speed_timeout` configuration I've set.

jacobgkau:
It seems like 1eb9567 has prevented the download from resuming after a network connection change-- starting on WiFi and plugging in Ethernet, the progress stops and doesn't start going up again.

mmstick:
Network changes do seem to give libcurl grief. I've enabled async-fetcher's own timeout configuration with a 5s timeout on a read from the HTTP server that doesn't progress.

jacobgkau:
Still able to recreate the timeout after closing/opening Settings during download and then plugging in Ethernet and unplugging it again, and the checksum was invalid again-- back to how things were in https://github.com/pop-os/upgrade/pull/261#issuecomment-1035442187.

When trying to recreate it a second time on the same boot (planning to skip the last set of network switches to avoid the timeout), I ended up having the progress in the GUI stall on WiFi after unplugging Ethernet; after closing and re-opening Settings, I had the green Upgrade button again, and clicking it made the button go away but didn't show the progress again:
Screenshot from 2022-02-10 16-02-22

To try and narrow down the corruption issue, I've tried downloads in the CLI with only manual cancel/resumes and only automatic network reconnections (the latter gave me one `failed to download ISO` error immediately after unplugging Ethernet, which I had to manually resume from); both of those downloads completed and the checksum was valid.

Next, I rebooted, opened the GUI, and tried one with just network disconnects/reconnects. I was also able to recreate an immediate timeout after one of the network switches, and had another immediate timeout when the download was about to finish, but no checksum failure after clicking the Upgrade button again after both of those.

Rebooted again, started the upgrade in the GUI on Ethernet, unplugged & waited for resume, closed Settings, opened it again, plugged Ethernet back in & let it finish-- got an immediate timeout at the end of the download, but no checksum mismatch.

@mmstick Are you able to recreate the timeout at the end of the download with the below steps?
1. Start upgrade in GUI on Ethernet.
2. Unplug & wait for resume via WiFi.
3. Close Settings, open it again.
4. Plug Ethernet back in, wait for download to finish (observe immediate timeout, but finishes successfully after clicking Upgrade again.)

And/or are you able to recreate the checksum mismatch with the below steps?

1. Start download in CLI on WiFi.
2. Cancel with Ctrl-C, then resume.
3. Plug in Ethernet, wait for progress to start going up again.
4. Cancel with Ctrl-C, then resume.
5. Unplug Ethernet, wait for progress to start going up again.
6. Cancel with Ctrl-C, then resume.
7. Cancel with Ctrl-C again.
8. Open the GUI, click Update.
9. Plug in Ethernet, wait for progress to start going up again.
10. Unplug Ethernet, wait for progress to start going up again.
11. Close/re-open Settings.
12. Plug in Ethernet again, wait for download to finish (observe immediate timeout, then MD5 mismatch after clicking Upgrade again.)

Seems like there are at least two problems: "timeouts" sometimes occurring in far less than 5 seconds (sometimes right at the end of the download), and the longer sequence resulting in a corrupted download for whatever reason (plus whatever GUI quirk that was which cased a non-working Update button.)

mmstick:
I've made a few changes that have resolved the issues.

- A successful fetch following a failure with reset the retry count for that file request
- libcurl does not have the ability to detect when the network connection it is using has closed, so I've added a network connection check that interrupts the fetch when network devices have changed, and waits for an online connection to proceed (up to 15s of waiting)
- The checksum validation was getting triggered whenever the fetcher returns a Fetched event, but async-fetcher was sending Fetched on fetch errors, so I'm now only sending Fetched events if the download was success

So I'm able to dynamically enable and disable Ethernet and WiFi and cancel the download while still getting a successful checksum validation on completion.

jacobgkau:
Recovery after network connection change is now usually much faster than before.

CLI:
- [X] Able to cancel/resume with appropriate output.
- [X] Download resumes when changing network connection.
- [X] Able to cancel/resume again after changing network connection.
- [X] Download resumes when changing network connection in the other direction.
- [X] Still able to cancel/resume with appropriate output.

GUI:
- [X] Able to resume.
- [X] Able to change network connection in both directions.
- [X] Closing the window continues downloading in the background; progress is reported again on re-opening.
- [ ] Able to change network connection in both directions again.

When following the above testing procedure, I re-opened the GUI window, plugged in Ethernet, waited for the download to mostly finish, and then unplugged Ethernet; the download stalled at that point.

Screenshot from 2022-02-11 10-06-50

After closing and re-opening Settings, I had the same issue as in https://github.com/pop-os/upgrade/pull/261#issuecomment-1035643199 where re-opening Settings showed the Upgrade button again, and clicking it didn't start/resume the download.

Screenshot from 2022-02-11 10-08-18
Screenshot from 2022-02-11 10-08-21

After manually restarting the `pop-upgrade` service, the download resumed in that GUI window without having to click anything else; however, the checksum was mismatched when it finished:

Screenshot from 2022-02-11 10-11-16

Maybe whatever caused that last stall-out/hang meant that the file didn't get a chance to be flushed.

mmstick:
I think the GUI may be hanging on the release check since it's also using libcurl and there's no network connection detection code in pop-upgrade. I can import the functionality here to catch that.

jacobgkau:
24f2329 hung when plugging Ethernet back in after re-opening the GUI (one step before where it hung in https://github.com/pop-os/upgrade/pull/261#issuecomment-1036429756, but similar in that it was a network reconnect after re-opening); still had the issue with the Update button disappearing without doing anything, still had a checksum mismatch.

mmstick:
I have something that works to resolve the GTK client issues. Which required gracefully handling service termination from systemd and altering the order that DBus events and DBus status are set to give time for the client to receive the last results before the service is terminated. I'll have the commits up soon.

mmstick:
Found issue with package updates. Calling `pop-upgrade release update`, canceling, then redoing it blocks the daemon.

mmstick:
Issue resolved

jacobgkau:
Cancelling on the CLI is back to saying `failed to download ISO` instead of `process has been cancelled`, is that expected? It was more clear when it said "cancelled."

Screenshot from 2022-02-16 16-05-09

mmstick:
Error message is now

```
Recovery upgrade status: recovery upgrade aborted: fetching from https://pop-iso.sfo2.cdn.digitaloceanspaces.com/21.10/amd64/nvidia/6/pop-os_21.10_amd64_nvidia_6.iso failed: task was canceled
```

jacobgkau:
In the CLI, it also looks like the final download progress doesn't get reported before the output moves on to calculating the checksum (I was confused why it was checksumming when the download wasn't finished yet):

Screenshot from 2022-02-17 10-09-34

(Ignore the `^[[15~^[[17~` in the output, I just hit the wrong key while trying to printscreen.)

jacobgkau:
<details>
<summary>d970e96 is failing to build:</summary>

```
  Downloaded 336 crates (54.9 MB) in 2.97s (largest was `windows` at 12.7 MB)
error: failed to sync

Caused by:
  found duplicate version of package `async-fetcher v0.8.0` vendored from two sources:

  	source 1: registry `crates-io`
  	source 2: https://github.com/pop-os/async-fetcher#1d5adf54
echo 'directory = "vendor"' >> .cargo/config
tar pcf vendor.tar vendor
rm -rf vendor
make[2]: Leaving directory '/home/system76/upgrade'
make[1]: Leaving directory '/home/system76/upgrade'
   dh_clean
 debian/rules build
dh build --with=systemd
   dh_update_autotools_config
   dh_autoreconf
   dh_auto_configure
   debian/rules override_dh_auto_build
make[1]: Entering directory '/home/system76/upgrade'
env CARGO_HOME="$(pwd)/target/cargo" make prefix=/usr
make[2]: Entering directory '/home/system76/upgrade'
rm -rf vendor; tar pxf vendor.tar
cargo build "--release" "--frozen" "--offline"
error: failed to resolve patches for `https://github.com/rust-lang/crates.io-index`

Caused by:
  failed to load source for dependency `sysfs-class`

Caused by:
  Unable to update https://github.com/pop-os/sysfs-class#ab63e7f6

Caused by:
  can't checkout from 'https://github.com/pop-os/sysfs-class': you are in the offline mode (--offline)
make[2]: *** [Makefile:75: target/release/pop-upgrade] Error 101
make[2]: Leaving directory '/home/system76/upgrade'
make[1]: *** [debian/rules:18: override_dh_auto_build] Error 2
make[1]: Leaving directory '/home/system76/upgrade'
make: *** [debian/rules:7: build] Error 2
dpkg-buildpackage: error: debian/rules build subprocess returned exit status 2
```
</details>

jacobgkau:
CLI:
- [X] Able to cancel/resume with appropriate output.
- [X] Download resumes when changing network connection.
- [X] Able to cancel/resume again after changing network connection.
- [ ] Download resumes when changing network connection in the other direction.
- [ ] Still able to cancel/resume with appropriate output.

GUI:
- [ ] Able to resume.
- [ ] Able to change network connection in both directions.
- [ ] Closing the window continues downloading in the background; progress is reported again on re-opening.
- [ ] Able to change network connection in both directions again.

fcec0f2 hung when attempting to switch network connection for the second time, from Ethernet to WiFi. Once it hung, Ctrl-C did not stop the process either.

Screenshot from 2022-02-21 13-23-09

After killing pop-upgrade and restarting it, clearing cache, and trying again, I was able to get farther, but hung the first time I tried switching from Ethernet to WiFi in the GUI:

CLI:
- [X] Able to cancel/resume with appropriate output.
- [X] Download resumes when changing network connection.
- [X] Able to cancel/resume again after changing network connection.
- [X] Download resumes when changing network connection in the other direction.
- [X] Still able to cancel/resume with appropriate output.

GUI:
- [X] Able to resume.
- [ ] Able to change network connection in both directions.
- [ ] Closing the window continues downloading in the background; progress is reported again on re-opening.
- [ ] Able to change network connection in both directions again.

Screenshot from 2022-02-21 13-30-23

Closing and re-opening the GUI at this point gave me the green Update button again, and upon clicking it, I get no progress again:

Screenshot from 2022-02-21 13-30-46

mmstick:
That's a big regression for such a small change. Guess I'll revert it.

mmstick:
Maybe this'll help. Added a shutdown to the concatenator task

jacobgkau:
CLI:
- [X] Able to cancel/resume with appropriate output.
- [X] Download resumes when changing network connection.
- [X] Able to cancel/resume again after changing network connection.
- [X] Download resumes when changing network connection in the other direction.
- [X] Still able to cancel/resume with appropriate output.

GUI:
- [X] Able to resume.
- [X] Able to change network connection in both directions.
- [X] Closing the window continues downloading in the background; progress is reported again on re-opening.
- [X] Able to change network connection in both directions again.

269c98c fixed the hangs, but still getting the checksum mismatch.

Screenshot from 2022-02-21 14-40-38

mmstick:
I have only one idea for the mismatch. I'm thinking it's possible that that a redownload begins to happen while the last fetch is still in the process of handling a shutdown. So I'll figure out a way to prohibit a resume until all spawned tasks are finished.

mmstick:
TimedOut errors may have been the cause. It was giving me issues with network changes causing a hang on a retry to a dead interface.

mmstick:
I've had two successful recovery upgrades with multiple disconnects and reconnects between WiFi and Ethernet, and combinations of connecting to both simultaneously and unplugging the other.

jacobgkau:
My testing still got a checksum mismatch:

Screenshot from 2022-02-21 19-26-00

Steps were the same I've been using:
- Start in CLI on WiFi.
- Cancel/Resume.
- Plug in Ethernet, wait for resume.
- Cancel/Resume.
- Unplug Ethernet, wait for resume.
- Cancel/Resume.
- Cancel.
- Open GUI, resume on WiFi.
- Plug in Ethernet, wait for resume.
- Unplug Ethernet, wait for resume.
- Close Settings, wait a few seconds; open Settings again, observe progress moved while it was closed and is still moving.
- Plug in Ethernet, wait for it to mostly finish (~200MB left).
- Unplug Ethernet, wait for resume, wait for finish.

mmstick:
Still getting successful recovery upgrades with these exact steps, but I'll keep trying.

mmstick:
How fast is your download? I guess it's possible for a race condition if the network is fast enough to download a part before a network-interrupted resume restarts.

mmstick:
Logs are fine. Behaving the same as my systems.

jacobgkau:
> How fast is your download? I guess it's possible for a race condition if the network is fast enough to download a part before a network-interrupted fetch could shut down.

I'm seeing ~80-100 MB/s on Ethernet, and ~5-8 MB/s on WiFi.

I've been testing on the darp7 with a USB-C Ethernet adapter, and confirmed the same issue occurs with the built-in Ethernet jack.

mmstick:
I hope this works. Added some logs to be able to compare when a fetch has flushed to when a subsequent fetch has begun. Made some further pruning of shutdown canceling. Testing it now.

mmstick:
Two successful recovery upgrades with the latest update. Although my Ethernet maxes out at 16 MB/s. There's some very verbose logging from async-fetching showing the order that parts are being fetched, concatenated, and flushed.

jacobgkau:
Here is a log from ec62690 ending with the checksum mismatch: [versbose-upgrade.txt](https://github.com/pop-os/upgrade/files/8120853/versbose-upgrade.txt)

mmstick:
According to the logs, no synchronization issues anywhere. Everything is systematically performing all operations concurrently in the same order. The recovery ISO completes a successful part concatenation and flush with each cancelation, time out, and network connection change before any subsequent request is started. Looking at other areas then...

jacobgkau:
I tried the same darp7 at the factory (significantly faster WiFi than mine, but slightly slower Ethernet), and it still had the checksum error.

I also tried with a gaze16 at the factory, and it did not have the checksum mismatch issue. Both machines had NVMe SSDs (the gaze16 had a Western Digital drive, while the darp7 has a Samsung drive.) I'll try to see if I can identify the difference on the darp7 causing it to fail more easily (hoping it's not just hardware.)

mmstick:
I'll push a small update with some potential fixes in a minute.

jacobgkau:
I can see that f9053a used larger part sizes (resumes tended to start farther back from the current position than they did before), but I still got the checksum error on darp7. Log in case it's useful: [1f9053a-darp7.txt](https://github.com/pop-os/upgrade/files/8133975/1f9053a-darp7.txt)

mmstick:
We could switch to a single connection download for the ISO and see if that works.

jacobgkau:
Ok, I recreated the checksum issue on the first try on lemp10 with a fresh 21.10 installation on a different SSD. lemp10 and darp7 are very similar internally, and the SSD was also a Samsung (darp7 has 980 Pro, lemp10 has 970 Evo Plus), but that at least rules out any software config on the darp7. (The darp7 install got its primary keyboard layout switched to a different language a while back, so I wanted to confirm that wasn't somehow causing this.)

mmstick:
That's a bummer. Switching to one connection in case that improves stability

mmstick:
Have to make some changes since single-connection downloads hasn't been tested yet

mmstick:
Had two successful recovery upgrades with a single connection now. There's less moving parts here so it should have a higher chance of success in the face of interruptions and cancelations.

jacobgkau:
Still got the checksum error on darp7 with 253dcb5. [253dcb5-darp7.txt](https://github.com/pop-os/upgrade/files/8136629/253dcb5-darp7.txt)

mmstick:
Perhaps a sync to disk before we get the length of the file between each retry.

jacobgkau:
No change: [7f10aee-darp7.txt](https://github.com/pop-os/upgrade/files/8137084/7f10aee-darp7.txt)

mmstick:
I'm out of ideas for now

mmstick:
Will try that

jacobgkau:
Here are logs with checksums from b6d88a1: [b6d88a1-darp7.txt](https://github.com/pop-os/upgrade/files/8163831/b6d88a1-darp7.txt)

mmstick:
Some issues at the moment with the transition. I'll notify when it's ready again.

mmstick:
Okay. The transition from using the isahc HTTP client to reqwest is ready for testing.

jacobgkau:
Still got the checksum error on 154dd2f: [154dd2f-darp7.txt](https://github.com/pop-os/upgrade/files/8164743/154dd2f-darp7.txt)

Comparing this log with the last one, I see that the concat checksums matched up to `1D88AAC5E6D859B7E5A1091728A7C34D`, then differed after the fourth (last) manual cancel on the earlier run. After that, the next concat from the earlier run was `7969B161E66449732D6A10C0BD7B2DDD`, which is not present in the later run. That last cancel would have been resumed from the GUI (with everything before then being performed on the CLI), that's the only unique thing about it I can think of compared with the earlier cancels.

mmstick:
Between cancels, have you noticed any cancels that had a gap or overlap in the bytes range?

```
get_many.rs:54: GET "/var/cache/pop-upgrade/recovery.iso" bytes=3137339392-3138715647
```

mmstick:
Seems to have resumed correctly
```
GET "/var/cache/pop-upgrade/recovery.iso" bytes=872415232-889192447
lib.rs:291: inner_request fetch result is Err(Canceled)
GET "/var/cache/pop-upgrade/recovery.iso" bytes=872415232-889192447
```

jacobgkau:
8c20466 still getting wrong checksum: [8c20466-darp7.txt](https://github.com/pop-os/upgrade/files/8165547/8c20466-darp7.txt)

jacobgkau:
Failure on 052f3e4: [052f3e4-darp7.txt](https://github.com/pop-os/upgrade/files/8171366/052f3e4-darp7.txt)

Comparing 052f3e4 to 154dd2f, the first 52 chunks (up to `ED8A84F04F1806F4CB9C6D7E15C62EEB`) had the same checksum, then 052f3e4 got `B33EFDB19456767A71C58CA24CE6B93E` which was not present in 154dd2f (so is presumably incorrect.) This one happened after the second-to-last cancel/resume, which would not have been after switching to the GUI, but just Ctrl-C and then running the command again in the terminal.

Interestingly, the four chunks 052f3e4 got during that period were all mismatching the other log, but then after resuming from the GUI, it got `A2BC4F171FB20360F3301B6CACF78A80`, which is present in the other log again. From there on, the rest of the chunks have consistent checksums with the earlier attempt, but the final checksum was mismatched. (Not sure if logging the checksum of the actual final file before or after concats would add any additional value over just checksumming the parts themselves.)

mmstick:
[log.txt](https://github.com/pop-os/upgrade/files/8172839/log.txt)

Here's a log of my own from another successful upgrade. I've increased data logged

jacobgkau:
Failure log from 5499263: [5499263-darp7.txt](https://github.com/pop-os/upgrade/files/8173650/5499263-darp7.txt)

Our checksums on the concat lines now already don't match on the first concat:

```
Mar 02 16:37:43 pop-os pop-upgrade[933]: [DEBUG] get_many.rs:55: RANGE /var/cache/pop-upgrade/recovery.iso:0 - bytes=0-16777215
Mar 02 16:37:43 pop-os pop-upgrade[933]: [DEBUG] get_many.rs:55: RANGE /var/cache/pop-upgrade/recovery.iso:1 - bytes=16777216-33554431
Mar 02 16:37:51 pop-os pop-upgrade[933]: [DEBUG] concatenator.rs:44: CONCAT /var/cache/pop-upgrade/recovery.iso:0 52D54746144D366C200EA41BBF48A724 16777216 bytes
```

```
[DEBUG] get_many.rs:55: RANGE /var/cache/pop-upgrade/recovery.iso:0 - bytes=33554432-50331647
[DEBUG] get_many.rs:55: RANGE /var/cache/pop-upgrade/recovery.iso:1 - bytes=50331648-67108863
[DEBUG] concatenator.rs:44: CONCAT /var/cache/pop-upgrade/recovery.iso:0 BE7163C1E37763F6084BF76015ED2958 16777216 bytes
```

Based on the ranges shown, it looks like yours was already partway done before the log started? I don't see the first few checksums from your log anywhere in mine, although the last few do match (which suggests to me that it's still the individual pieces' checksums being logged.)

mmstick:
Yeah the first few parts were already fetched

mmstick:
Yeah, this is strange. The fetched ranges from the darp7 are wrong for many of the fetches. At random points inbetween the checksums differ, and sometimes matches

mmstick:
Updated logs (full): 
[gaze15.txt](https://github.com/pop-os/upgrade/files/8173851/gaze15.txt)

jacobgkau:
Failure log from 29526df: [29526df-darp7.txt](https://github.com/pop-os/upgrade/files/8188816/29526df-darp7.txt)

mmstick:
Closing this for now

PR #262, opened 2022-02-10
i18n(it): Add Italian translation

matilde-t:


jacobgkau (dismissed):
Thank you for the translation! This shows up properly in a user account set to Italian.

mmstick (approved):

matilde-t:
I suddenly remembered that in Italian we don't use title case, sorry I fixed that

jacobgkau (approved):
Still displays as expected.

PR #269, opened 2022-03-02
fix: Replace pop-os-ppa.sources with pop-os-release.sources

### LINKED ISSUE #268: Duplicate APT sources force upgrading clean install of 21.10 to 22.04
`pop-upgrade release upgrade -f` on a clean install of 21.10 results in two files with duplicate apt sources, causing errors.

The clean install of 21.10 has a file `/etc/apt/sources.list.d/pop-os-release.sources`, while pop-upgrade creates this as `/etc/apt/sources.list.d/pop-os-ppa.source` (which exists in previous releases, and installs upgraded from a previous release).

`pop-upgrade` should probably remove `pop-os-ppa.source` if it exists, and create `pop-os-release.sources`. Assuming we don't need to preserve any user changes to this particular file.

mmstick:
Closes #268


ids1024 (dismissed):

jacobgkau:
I noticed while testing this that pop-upgrade is using `22.04` as the codename (as opposed to `jammy`), so it's still failing before it starts downloading the new release.

image

Can this be fixed? If not, I can still regression test this with older releases, but it won't do a lot of good yet.

mmstick:
Yeah, I'll get that ready

mmstick:
Possibly related to this https://bugs.launchpad.net/ubuntu/+source/dbus/+bug/1962036

jacobgkau (approved):
What this PR is adding (handling of the transition from `pop-os-ppa.sources` to `pop-os-release.sources`) is working at this point, so I'm approving this.

I've tested the following:
- [X] 21.10 (with `pop-os-release.sources`) -> 22.04: fails to upgrade, but the apt sources are correct.
- [X] 21.10 (with `pop-os-ppa.sources`): running `pop-upgrade release repair` does not rename the file. Running `pop-upgrade release upgrade -f` moves the old file to `pop-os-ppa.sources.save` and creates the new file as `pop-os-release.sources`.
- [X] 18.04 (with `system76-ubuntu-pop-bionic.list`)-> 20.04: focal PPA ends up in `sources.list` after the upgrade. (My VM had trouble performing the offline portion of the upgrade that I think is related to Plymouth; it also failed without this branch, so it's not a regression.)
- [X] 20.04 (with `sources.list`) -> 21.10: only `pop-os-release.sources` exists after the upgrade.
- [X] 20.04 (with `system76-ubuntu-pop-focal.list`) -> 21.10: only `pop-os-release.sources` exists after the upgrade.
- [ ] 21.04 (with `system76-ubuntu-pop-hirsute.list`)-> 21.10: unable to test because hirsuite builds aren't happening anymore (which also means this update won't go to hirsute, anyway.)

PR #225, opened 2021-11-16
Schedule Automatic Updates

mmstick:
Requires `pop-system-updater` to be installed.

Requires https://github.com/pop-os/system-updater to be released.

Requires a patch to pop-shop to disable its own notifications

Closes #217
Closes #273 
Requires https://github.com/pop-os/system-updater/pull/1

PR #272, opened 2022-03-21
chore: Update dependencies and refactor various parts of the service and GUI

mmstick:
This is in preparation for integrating with system-updater and resumable download support PRs.


jacobgkau:
In a 20.04 virtual machine, trying to upgrade in a GUI, I got a hang while downloading the first package (on the "upgrading current release" step). The VM's network indicator looked like it was downloading something, but it never finished.

Screenshot from 2022-03-21 14-30-05

Looking in `/var/cache/apt/archives/`, I see a `.part0` file for the package in question continuously being created and removed; the file itself (the one without `.part0` never breaks 200K large before resetting back to 0.

Screenshot from 2022-03-21 14-34-28

Trying again with the CLI, I'm seeing the same issue occur.

image

jacobgkau:
- Virtual machine 20.04 -> 21.10 w/ problem package list (CLI): successful.
    - Fails to install updates for current release if /boot partition is full and dpkg configure returns an error; not a regression.
    - Once on 21.10, apt upgrade tries to install chromium snap unless pop-transition is run first; not a regression.
- Physical machine 20.04 -> 21.10 w/ non-default packages (GUI): ran into an issue where the GUI got stuck on checking for updates:

Screenshot from 2022-03-22 14-20-02

After closing and re-opening settings, I just got a spinner:

Screenshot from 2022-03-22 14-21-18

I tried running `sudo killall pop-upgrade` and `sudo systemctl restart pop-upgrade`, and systemctl hung on trying to restart the service:

Screenshot from 2022-03-22 14-22-24

Eventually, it timed out:

Screenshot from 2022-03-22 14-24-09

Once that was finished, opening the GUI worked again; however, I got it to hang on `Checking for updates ...` again after closing and re-opening it two more times.

mmstick:
I'll purge 208be29 because that's most likely the cause.

jacobgkau:
I'm no longer getting hangs at that point.

On the oryp8 trying to upgrade from 20.04 using the GUI, I got a `Looks like we've encountered an issue!` error message after the current release was upgraded (but before the new release was downloaded). Here is the journalctl output: [oryp8-0eaa0cb.txt](https://github.com/pop-os/upgrade/files/8328612/oryp8-0eaa0cb.txt) 

Also, it didn't look like the recovery partition updated before the release upgrade started like I expected it to. After I dismissed the error message from the release upgrade failing, the GUI still showed a recovery partition update available, but the `Update` button was still greyed out as if the release upgrade was still in progress:

Screenshot from 2022-03-22 18-11-23

jacobgkau:
oryp7 (different install) 20.04 -> 21.10 on CLI had some issues this morning.

- Recovery download output didn't look like it was finished when it was synced to the recovery partition
- Release upgrade failed, and the reboot message showed anyway

Screenshot from 2022-03-23 09-41-00

Here is the log: [oryp7-22f3683.txt](https://github.com/pop-os/upgrade/files/8334421/oryp7-22f3683.txt) but I'm actually still getting output from pop-upgrade in journalctl (the current release upgrade is still running), so it seems like there was a disconnect between the CLI interface and the daemon.

jacobgkau:
Back in a VM, I got some strange output when one of the packages for the new release failed to download:

Screenshot from 2022-03-23 13-24-54

I see the same package listed three times before the error message. I'm not sure the "fetched" counter should be going up if it's retrying the same package-- not sure if it actually means what it looks like, but it at least looks wrong.

When I tried again, the packages for the new release did finish downloading, and the upgrade completed successfully.

jacobgkau:
oryp8 on 20.04 is still getting a `Looks like we've encountered an issue!` error message in the GUI trying to upgrade to 21.10:
[oryp8-66ed872.txt](https://github.com/pop-os/upgrade/files/8352502/oryp8-66ed872.txt)

Once again, I still have packages being fetched as I type this, after the GUI errored out.

mmstick:
At what point in the log did that message appear?

jacobgkau:
> At what point in the log did that message appear?

After rebooting manually (which did not start the offline upgrader), I tried to reset things by clearing all packages with `21.10` in the name from `/var/cache/apt/archives` and switching my apt sources & preferences back to Focal. This time, instead of the error message, I got an `Upgrade Ready` notification and a green `Upgrade` button before things had finished downloading. Here is a screenshot of the issue occurring side-by-side with the log:

https://user-images.githubusercontent.com/7199422/160172875-433eca89-1b65-4a39-b71f-c6dcd9193ce5.mp4

It happened while the log was sitting on `[INFO] daemon/mod.rs:796: release upgrade: updating package lists`.

mmstick:
I've determined the problem and found the solution. Pushing patches in a minute.

mmstick:
Currently testing

mmstick:
Installs successfully with CLI without stalling at the end

jacobgkau:
- 20.04 -> 21.10 BIOS VM w/ problem package list via CLI: successful
- 20.04 -> 21.10 UEFI physical machine w/ non-default packages via GUI: successful
- 21.10 -> 22.04 UEFI physical machine w/ non-default packages via CLI: ran into the issue where the CLI exited before the downloads started again (this is on 651b025).
Screenshot from 2022-03-29 10-58-02

mmstick:
It seems to be in a good state now.

jacobgkau (changes requested):
20.04 -> 21.10 BIOS VM w/ problem package list via CLI: 
20.04 -> 21.10 UEFI physical machine w/ non-default packages via GUI: Once again had the GUI notify me that the upgrade was ready & give me a green download button before the packages had started downloading.

Screenshot from 2022-03-31 11-25-07

mmstick:
Yeah, I'm about to start GTK testing again. I've only just got the CLI client working well.

mmstick:
I'll have some commits pushed soon which resolve the GTK issues, and further improve handling of hyper connection errors.

mmstick:
I have successfully performed a release upgrade from 20.04 to 21.10 with no errors between the initial package updates, recovery ISO upgrade, and release package fetch.

jacobgkau (changes requested):
[8d83a80](https://github.com/pop-os/upgrade/pull/272/commits/8d83a80dab0317315b222b489ced9001cb0b6ae9) still has the GUI reporting finished before packages are downloaded issue.

Screenshot from 2022-03-31 21-02-56

mmstick:
Have to test changes from the CI repository, unfortunately. Locally-installed packages get auto-downgraded before an upgrade is started.

mmstick:
The GTK client's DBus interactions are currently being done in a background thread isolated from the GUI code. If we move that code into a new binary and use IPC between the GTK client and this theoretical new binary, a need to potentially restart the GTK application wouldn't be necessary anymore.

jacobgkau:
- 20.04 -> 21.10 BIOS VM w/ problem package list via CLI: successful
- 20.04 -> 21.10 UEFI physical machine w/ non-default packages via GUI: successful
- 21.10 -> 22.04 BIOS VM w/ problem package list via CLI: still in progress
- 21.10 -> 22.04 UEFI physical machine w/ non-default packages via CLI: successful

There is a small regression with CLI output. Before:

image

After:

Screenshot from 2022-04-04 13-46-45

Notice that the new `Fetch Complete` message does not erase the download progress that it's trying to replace, so the end of the download progress still shows up after it, which looks broken. (I would think it'd be better to show `Fetch Complete` on a new line, I don't know what benefit hiding the size of the fetch from the user would have.)

mmstick:
That's definitely an issue

jacobgkau (approved):
CLI output for recovery update is now working normally.

- 20.04 -> 21.10 BIOS VM w/ problem package list via CLI: successful
- 20.04 -> 21.10 UEFI physical machine w/ non-default packages via GUI: successful
- 21.10 -> 22.04 BIOS VM w/ problem package list via CLI: successful
- 21.10 -> 22.04 UEFI physical machine w/ non-default packages via CLI: successful
- On the physical machine, initiated a refresh install from the pop-upgrade GUI: successful

Once this has engineering approval, can we merge this, then change the base branch of https://github.com/pop-os/upgrade/pull/225 to master and make sure it's rebased/up-to-date?

jackpot51 (approved):

PR #274, opened 2022-04-07
fix: Revert to 86da11d

jacobgkau:
I did this with `git revert --no-commit 86da11d..HEAD` followed by a commit, and validated with `git diff 86da11d` giving no output. If someone from engineering could make sure this was done correctly, that would be great.

(I first tried just branching from 86da11d, but GitHub wouldn't let me open a PR because there were no changes to compare, I'm guessing because master already contains that commit-- we'd need to force-push master to do it that way, which would be bad form. Doing it this way will require another repo-release update to whatever the new commit is once this is merged.)


mmstick:
I wouldn't mind doing a force push to the right commit, and then the system-updater update would contain all the commits that were reverted.

jacobgkau:
That sounds like a good plan if you'd rather do it that way.

PR #277, opened 2022-04-14
22.04 release notes

Ampersandstorm:
This PR adds the "22.04" change log.

### DIFF
--- a/changelogs/22.04
+++ b/changelogs/22.04
@@ -2,6 +2,8 @@
 
 * COSMIC
 ** GNOME 42 support has been added for COSMIC.
+** The Shell Theme for GNOME 42 has been updated.
+** Icons are now SVG-based instead of PNG-based.
 ** Dark and light modes can have independent backgrounds.
 ** Settings for Desktop Options, Background, Appearance, Dock, and Workspaces are now searchable in Launcher.
 ** The Pop robot face has been added as the default avatar.
--- a/changelogs/22.04
+++ b/changelogs/22.04
@@ -31,7 +31,6 @@
 ** PipeWire replaces the PulseAudio audio system.
 ** Mesa 22 has been added.
 ** Irqbalance has been removed from the default installation.
-** Rust is now packaged in the default installation, and will be kept up-to-date with upstream releases.
 
 * System Improvements
 ** The journald log max disk capacity is now limited to 1GB.
--- a/changelogs/22.04
+++ b/changelogs/22.04
@@ -1,0 +1,50 @@
+* Pop!_OS is now based on Ubuntu 22.04.
+
+* COSMIC
+** GNOME 42 support has been added for COSMIC.
+** Dark and light modes can have independent backgrounds.
+** Settings for Desktop Options, Background, Appearance, Dock, and Workspaces are now searchable in Launcher.
+** The Pop robot face has been added as the default avatar.
+** Improvements to cosmic-workspaces:
+    ** An issue with the location of the Workspace picker on secondary displays has been fixed.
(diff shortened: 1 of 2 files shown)


mmstick:
[Needs to also reference the file here](https://github.com/pop-os/upgrade/blob/master/src/changelogs.rs)

changelogs/22.04
```diff
@@ -0,0 +1,52 @@
+* Pop!_OS is now based on Ubuntu 22.04.
+
+* COSMIC
+** GNOME 42 support has been added for COSMIC.
+** The Shell Theme for GNOME 42 has been updated.
+** Icons are now SVG-based instead of PNG-based.
+** Dark and light modes can have independent backgrounds.
+** Settings for Desktop Options, Background, Appearance, Dock, and Workspaces are now searchable in Launcher.
```
jacobgkau, line 8: I would probably say "searchable in the Launcher."

changelogs/22.04
```diff
@@ -0,0 +1,52 @@
...
+** Improvements to cosmic-workspaces:
+    ** An issue with the location of the Workspace picker on secondary displays has been fixed.
+    ** An issue with the scaling of the Workspace picker on displays of varying resolutions has been fixed.
+    ** Performance has been improved.
+    ** An issue with alignment on HiDPI displays has been fixed.
+    ** The OSD picker is now smaller.
+** COSMIC logging has been limited, resulting in increased system performance.
+
+* Pop!_OS System Updater
+** New upgrade and options for installing upgrades are now available in Settings > OS Upgrade and Recovery.
```
jacobgkau, line 19: I didn't get what "New upgrade and upgrade installation options" meant the first time, but I see with this rewording that the entire point is referring to the auto update feature. I would say "options for installing updates" since we generally use "upgrade" to refer to release upgrades and "update" to refer to installing updates on the current release. You can probably just say "New options for installing updates" and not have the word "upgrades" in this line at all (I still don't understand why the word "upgrade" is being used twice in this bullet point as-is.)

changelogs/22.04
```diff
@@ -0,0 +1,52 @@
...
+    ** To reduce distractions, update notifications are now set to show weekly and can be set to daily or monthly.
+    ** Upgrades can be scheduled and installed automatically (includes Debian, Flatpak, and *nix packages).
```
jacobgkau, line 21: Again, we might want to stay consistent with calling them "updates" instead of "upgrades" for this feature, but not necessarily an issue if we don't want to go that direction.

jacobgkau (dismissed):
Changelog looks good to me. This PR does not cause 22.04 to be reported as available.

changelogs/22.04
```diff
@@ -0,0 +1,52 @@
...
+* Pop!_Shop
+** Installed NVIDIA drivers are now visible in the Pop!_Shop and older drivers are available to install.
+** The Pop!_Shop is now uninstallable, and the GNOME Software Center can be installed.
+
+* Packaged Applications
+** Debian packaging has been added for Firefox.
+** PipeWire replaces the PulseAudio audio system.
+** Mesa 22 has been added.
+** Irqbalance has been removed from the default installation.
+** Rust is now packaged in the default installation, and will be kept up-to-date with upstream releases.
```
mmstick, line 34: This one should be `Rust is now packaged in the Pop repositories`, since it's not part of any install.

mmstick: Or left out entirely since it's only relevant to Rust programmers.

jacobgkau (approved):

mmstick (approved):

PR #286, opened 2022-04-27
Deferred Improvements

mmstick:
- Relies on upgrade branch at https://github.com/pop-os/system-updater/pull/6
- Unholds all packages that might be held before setting up a release upgrade
- Removes two potentially-problematic packages (update-notifier-common and nodejs)
- Performs an autoremove when removing packages
- Holds snap packages after fetching
- 2 concurrent package fetches with 0ms delay between requests
- GTK panel is scrollable with https://github.com/pop-os/gnome-control-center/pull/227

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,7 +95,7 @@ dependencies = [
 [[package]]
 name = "async-cron-scheduler"
 version = "1.0.0"
-source = "git+https://github.com/pop-os/system-updater?branch=master#36436e2a98c87df6b0f5f4ffd41b51191ba02d39"
+source = "git+https://github.com/pop-os/system-updater?branch=upgrade#afe4fac791b2122a0128c6fd19568d5f11525334"
 dependencies = [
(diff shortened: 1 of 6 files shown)


jackpot51 (approved):

jackpot51 (approved):

n3m0-22 (approved):
Tested upgrading 20.04 and 21.10. Both were successful with no regressions.

n3m0-22:
@mmstick 
Update: I double checked upgrading from 20.04 and 21.10. All snap packages are left alone. Nodejs is gone. I installed the old Chromium from [here](https://launchpad.net/~system76/+archive/ubuntu/pop) on 20.04 and it offered me a new Flatpak after upgrading. I installed handbrake since it comes from Ubuntu and it is not a snap package after upgrading. I did notice that update-notifier-common is still installed after upgrade on both this time. Otherwise everything checks out.

PR #289, opened 2022-04-29
Updates

mmstick:
- Removes ttf-mscorefonts-installer on upgrade because it causes some conflicts.
- Updates the system-updater GTK widget, which has a new translation

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,9 +37,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.56"
+version = "1.0.57"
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jackpot51:
@mmstick any chance we could have pop-upgrade use the system-upgrader FFI so we don't have to keep doing cargo.lock changes?

n3m0-22 (approved):
Upgraded from 21.10 to 22.04 successful and all packages that should be were removed including `ttf-mscorefonts-installer`.

mmstick:
@jackpot51 I'll see about making an FFI interface for it so we can do that.

PR #293, opened 2022-05-02
Auto-remove remoteless packages before upgrade

mmstick:
This will auto-remove locally-installed packages with no remote repository. These packages either belonged to a PPA or were installed with Eddy / dpkg by hand, which might conflict with apt's dependency resolution in the future. You can get a list of packages that would be removed by running this:

```
git clone https://github.com/pop-os/apt-cmd
cd apt-cmd
cargo run --example remoteless
```


jacobgkau (changes requested):
This does not install on 20.04. I get this output when trying to upgrade pop-upgrade manually:

```
system76@pop-os:~$ sudo apt install pop-upgrade=1.0.0~1651672751~20.04~22af00a
Reading package lists... Done
Building dependency tree       
The following packages were automatically installed and are no longer required:
  libgsound0 librygel-core-2.6-2 librygel-db-2.6-2 librygel-renderer-2.6-2 librygel-server-2.6-2 libs76-hidpi-widget
  libwhoopsie-preferences0 libwhoopsie0 rygel whoopsie-preferences
Use 'sudo apt autoremove' to remove them.
The following additional packages will be installed:
  pop-system-updater
The following packages will be REMOVED:
  gnome-control-center libpop-upgrade-gtk pop-desktop system76-driver system76-driver-nvidia
  pop-system-updater
The following packages will be upgraded:
  pop-upgrade
WARNING: The following essential packages will be removed.
This should NOT be done unless you know exactly what you are doing!
  pop-desktop gnome-control-center (due to pop-desktop)
1 upgraded, 1 newly installed, 5 to remove and 404 not upgraded.
Need to get 3,605 kB of archives.
After this operation, 351 kB of additional disk space will be used.
This operation is not permitted because it will break the system.
Abort.
```

Still trying to figure out why this is happening. It installs and seems to be working as expected on 21.10.

mmstick:
Sounds like a conflict with the pop-system-updater package perhaps?

jacobgkau:
> Sounds like a conflict with the pop-system-updater package perhaps?

I installed other updates to make sure it wasn't anything else I have installed, and pop-system-updater is able to be installed. I then get this when running the same command as above:

```
system76@pop-os:~$ sudo apt install pop-upgrade
Reading package lists... Done
Building dependency tree       
The following packages were automatically installed and are no longer required:
  libgsound0 librygel-core-2.6-2 librygel-db-2.6-2 librygel-renderer-2.6-2
  librygel-server-2.6-2 libs76-hidpi-widget libwhoopsie-preferences0
  libwhoopsie0 rygel whoopsie-preferences
Use 'sudo apt autoremove' to remove them.
The following packages will be REMOVED:
  gnome-control-center libpop-upgrade-gtk pop-desktop system76-driver
  system76-driver-nvidia
The following packages will be upgraded:
  pop-upgrade
WARNING: The following essential packages will be removed.
This should NOT be done unless you know exactly what you are doing!
  pop-desktop gnome-control-center (due to pop-desktop)
1 upgraded, 0 newly installed, 5 to remove and 8 not upgraded.
Need to get 1,946 kB of archives.
After this operation, 11.6 MB disk space will be freed.
This operation is not permitted because it will break the system.
Abort.
```

Which doesn't mention system-updater, so maybe it's something else.

jacobgkau:
The `libpop-upgrade-gtk` from this staging branch depends on `libpop-system-updater-gtk`, but that package has no sources available on this 20.04 machine, that's why it's trying to remove `libpop-upgrade-gtk`.

mmstick:
I'll get that fixed then

mmstick:
The newer version of pop-system-updater will get published there soon.

jacobgkau:
This doesn't seem to cause any problems, although it doesn't work very consistently across a variety of situations.

- Tested with davinci-resolve installed from a .deb (with no repo/PPA); after the upgrade, it was no longer installed (it was removed prior to the offline portion of the upgrade.)
- Tested with qBittorrent installed from a PPA, with the package name being the same as Ubuntu's own package; after the upgrade, it was still installed at the PPA's version.
    - The PPA's version was _not_ removed by this new mechanism, because the package still had a remote (Ubuntu); additionally, it was not downgraded to Ubuntu's Impish version while upgrading packages for the current release, presumably because the version number from the PPA was higher than the version number from Ubuntu. It was also not downgraded to Jammy's version during the release upgrade; after the upgrade, its candidate version was still the version from the PPA (which was no longer enabled).
    ```
    system76@pop-os:~$ apt policy qbittorrent
    qbittorrent:
      Installed: 1:4.4.2.99~202205021033-7549-0c8220c9f~ubuntu21.10.1
      Candidate: 1:4.4.2.99~202205021033-7549-0c8220c9f~ubuntu21.10.1
      Version table:
     *** 1:4.4.2.99~202205021033-7549-0c8220c9f~ubuntu21.10.1 100
            100 /var/lib/dpkg/status
         4.4.1-2 500
            500 http://us.archive.ubuntu.com/ubuntu jammy/universe amd64 Packages
    ```
- Tested with Spotify installed from a PPA.
    - On an upgrade from 20.04, the PPA was removed, but the `spotify-client` package was not removed (because `spotify-client` was also available from the Pop proprietary PPA on 20.04.). It's still installed and now orphaned after the upgrade, since Pop 22.04 doesn't include that package (although we do have pop-transition to handle that specific app):
    ```
    system76@pop-os:~$ apt policy spotify-client
    spotify-client:
      Installed: 1:1.1.84.716.gc5f8b819
      Candidate: 1:1.1.84.716.gc5f8b819
      Version table:
     *** 1:1.1.84.716.gc5f8b819 100
            100 /var/lib/dpkg/status
    ```
     - On an upgrade from 21.10, the PPA was removed and the `spotify-client` package was also removed, because we weren't providing Spotify in the proprietary PPA for 21.10. (This sidesteps pop-transition.)

Thoughts on the inconsistency? Is removing only completely-orphaned packages accomplishing the goal, or should this be more robust? I'm not aware of the specific case/cases that prompted this feature.

mmstick:
So it isn't a good idea to remove sourceless packages after fetching package lists for the `next` release, because some of those are essential packages to the `current` release. It'd only be safe to do that after the release is finished upgrading.

I was able to locate and downgrade packages that are installed locally with a package version higher than from the sources. I was able to get spotify-client to automatically uninstall, and qbittorrent is correctly downgraded to the repository version after being installed from a PPA.

jacobgkau:
spotify-client is still removed when upgrading from 21.10 to 22.04 and not removed when upgrading from 20.04 to 22.04. This isn't a problem since pop-transition is there to handle it; pop-transition shows up for it now since the installed version's source was Pop.

Still need to do a little more regression testing on this.

jacobgkau (approved):
I've confirmed the qBittorrent case is also working as it should now (ends up on the Ubuntu Jammy version after the upgrade). Standalone non-PPA packages like Resolve still get removed. Regression testing with a variety of problem packages & the cursed VM all passed.

jackpot51 (approved):

Derisis13:
I have read trough the conversation and the commit messages but I still can't tell what is the conclusion: do I need to create a list of the applications I have installed and have no sources before upgrading or are they excluded from the removal automatically (by generating a list of them before any upgrading takes place)?

mmstick:
This is something that is performed automatically by these changes. There is no user intervention required. This change has not been released yet.

PR #303, opened 2022-06-22
i18n(cs): Add Czech translation

YoriMirus:
Hello, I have added a Czech translation according to the instructions found here: 
https://www.reddit.com/r/pop_os/comments/nb2rly/list_of_translatable_pop_os_projects/

In Czech update and upgrade are the same word, so I hope that won't cause confusion.


n3m0-22 (approved):
This looks good! Thanks for your contributions.

PR #315, opened 2022-10-05
Update dependencies and resolve recovery partition full error

mmstick:
Similar to other PRs, clap update potentially changes behavior of CLI arguments.
Also should resolve Dependabot security alerts.


n3m0-22 (changes requested):
I'm seeing this error when running `pop-upgrade recovery upgrade from-release`.

```
thread 'main' panicked at 'ArgAction::SetTrue / ArgAction::SetFalse is defaulted', /build/pop-upgrade-sW5Ews/pop-upgrade-1.0.0~1664967067~22.04~26e9b05/vendor/clap/src/parser/matches/arg_matches.rs:176:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

n3m0-22 (changes requested):
I can't seem to reboot into the recovery partition using the CLI or GTK. Using the spacebar at boot is the only way to boot to recovery.

PR #330, opened 2024-05-10
Prepare for 24.04 upgrades

mmstick:
- [x] Upgrade all dependencies, and the dependencies of our dependencies
- [x] Enable development flag upgrades to 24.04
- [x] Make necessary adjustments to install scripts


mmstick:
This will require that the updates currently in master_noble are released.

jackpot51 (dismissed):

jacobgkau:
Is this `no build available` error when forcing an upgrade normal? I don't remember that from testing previous releases pre-release.

screenshot-2024-05-16-17-04-00

The upgrade did continue, despite the message saying it was aborted.

jacobgkau:
Ah, I see, the "no build available" thing is because there are no 24.04 published isos yet.

Updates for noble have been released now. After adding this branch in the 22.04 VM again, applying all updates, rebooting, and then running `pop-upgrade release upgrade -f`, this time it seemed to think it was ready to reboot way earlier than it should have (it didn't do this when I tried last week):

screenshot-2024-05-20-21-05-43

The log doesn't look like it contains anything super useful (this was collected after rebooting): [out.txt](https://github.com/pop-os/upgrade/files/15381854/out.txt)

Any idea why this is happening? I replicated it a second time, making sure there weren't any updates available (including for `update-manager-core` and `ubuntu-release-upgrader-core`). pop-upgrade still just downgraded those to the released versions (presumably because they weren't pinned and it reset the system's repos) and then told me to reboot.

screenshot-2024-05-20-21-35-58

I think those packages not being pinned during upgrades (and thus showing up in the output when testing pop-upgrade branches) was existing behavior, but the failure's new.

XV-02:
I have found, and continue to find, that I need to run the upgrade command (`pop-upgrade release upgrade -f`) two to three times to actually kick an upgrade off. [I added a few lines to the output for clarity](https://github.com/user-attachments/files/16167549/pop-upgrade.log) but this shows a few reported errors. 

On the first attempt, the release upgrade was aborted. On the second, the release upgrade kicked off, but the recovery upgrade was aborted.

mmstick:
Do you have logs from the pop-upgrade daemon?

mmstick:
May need to use `env S76_TEST=1` to prevent it from upgrading to the version in the master repository.

XV-02:
What would the expected behaviour be on a system which does not have Cosmic installed/configured before the upgrade? It look like the boot process after running upgrade hangs while attempted to start gdm?

mmstick:
It transitions to what is defined in the [pop-desktop packaging for 24.04](https://github.com/pop-os/desktop/blob/master_noble/debian/control). The pop-de-cosmic metapackage replaces pop-de-gnome, so most gnome packages will be automatically removed on upgrade. Since cosmic-session provides cosmic-greeter, cosmic-greeter should be set up instead of gdm.

jacobgkau (changes requested):
VirtualBox test:

1. Fresh install of 22.04 Intel/AMD (iso 42)
    - Snapshot so it can be rolled back to later.
2. Add the `pop-upgrade` branch and do `sudo apt update` and `sudo apt upgrade`
3. Run `S76_TEST=1 pop-upgrade release upgrade -f`

With this setup, there were no issues before rebooting into the offline upgrader. After the offline upgrade, the system got stuck at `Upgrade complete. Now rebooting...`. This may just be a VM quirk?

After manually resetting, I got stuck in the boot at `Started gdm.service - GNOME Display Manager`. I saw this on my last hardware upgrade as well.

screenshot-2024-08-08-22-11-55

`/etc/systemd/system/display-manager.service` still points towards `gdm.service`. Running `sudo systemctl disable --now gdm.service` followed by `sudo systemctl enable --now cosmic-greeter.service` results in COSMIC Greeter starting successfully. The upgrader needs to handle this step.

jacobgkau (commented):
Great news! With the latest commit, an upgrade of a fresh VM install completes successfully, and is properly switched over to use cosmic-greeter.

Is there a reason why we leave gdm3 (and gnome-shell, for that matter) installed? Do we want to try and add an additional post-upgrade step to remove some of the legacy GNOME packages? If we're leaving it there for safety, that's fine (although it's not being tested). Thought we might want to think about it now, since we generally can't have packages be removed in the middle of an existing release should they start causing problems.

Also, there was one other issue. Before downloading packages, `pop-upgrade` successfully downloaded the new recovery ISO (now that we have alpha ISOs published), but it failed to install it:

screenshot-2024-08-09-21-23-13

After the upgrade's complete, booting into recovery still gives me a 22.04 recovery image. We definitely don't want that.

jacobgkau:
Here's what's in journalctl when the iso failed to install to the recovery partition, in case that helps determine the issue.

screenshot-2024-08-09-22-53-43

jacobgkau:
I'm seeing the same recovery partition issue from https://github.com/pop-os/upgrade/pull/330#pullrequestreview-2231178063 testing again on a physical machine today.

jacobgkau:
A physical machine (lemp13) still got stuck at `Upgrade complete. Now rebooting...`



Pressing `Esc` shows this.



After holding the power button down to reboot, the system did present COSMIC Greeter and log into COSMIC successfully.

I noticed that while `gnome-shell` and `gdm3` are still installed, COSMIC Greeter doesn't show GNOME as an option; the only item in the gear menu is COSMIC. So if we're leaving those installed on purpose as a fallback for existing users, the current setup doesn't really work for that.

jacobgkau:
After doing a fresh 22.04 installation on the lemp13, installing all updates (with this branch), rebooting, and attempting to upgrade, I got an extremely early exit erroneously showing the "reboot now" message in the middle of fetching the recovery iso.

Screenshot from 2024-09-23 09-35-22

Here are the logs for this boot (before running the upgrade with `-f`, I first tested running the command without that, and opening the Upgrade settings page): [pop-upgrade-fail.txt](https://github.com/user-attachments/files/17100271/pop-upgrade-fail.txt)

@mmstick Can you tell what caused this? Looking at the logs, I see the recovery download stopped when systemd reported it had finished checking for an available upgrade.

mmstick:
The error in the log is from the pop-upgrade daemon:

```
Sep 23 09:34:54 pop-os pop-upgrade[3714]: [INFO ] daemon/mod.rs:1032: performing a release check
Sep 23 09:34:54 pop-os pop-upgrade[3714]: [INFO ] daemon/mod.rs:1039: Release { current: "22.04", lts: "true",  next: "24.04", available: false }
Sep 23 09:34:54 pop-os pop-upgrade[3714]: [INFO ] release_api.rs:58: checking for build 22.04 in channel nvidia
Sep 23 09:34:55 pop-os pop-upgrade[3714]: thread 'tokio-runtime-worker' panicked at /build/pop-upgrade-WYcn4N/pop-upgrade-1.0.0~1723212513~22.04~7100826/vendor/async-fetcher/src/get.rs:94:10:
Sep 23 09:34:55 pop-os pop-upgrade[3714]: called `Result::unwrap()` on an `Err` value: JoinError::Cancelled(Id(117))
Sep 23 09:34:55 pop-os pop-upgrade[3714]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

mmstick:
When attempting to upgrade, did you set `env S76_TEST=1`?

jacobgkau:
I'm able to make that happen pretty easily by starting an upgrade within a few minutes of rebooting. I just made it happen again on master using `pop-upgrade recovery upgrade from-release`, so I'll just open an issue about it.

jacobgkau:
Alright, @mmstick I've narrowed down exactly where `upgrade.sh` is getting stuck after the upgrade is finished. It's getting stuck on the `systemctl unmask acpid pop-upgrade` command.

jacobgkau:
I noticed that the master branch currently isn't built for Noble. I'm assuming the reason it was getting stuck may have been because there was no longer a pop-upgrade service to unmask? (Running it in a terminal just gives an error message instead of hanging, though. It shouldn't just sit there, no matter what.)

I'll test the fix soon.

jacobgkau:
The notifier issue appears to be resolved. When `pop-upgrade-notify.timer` starts `pop-upgrade-notify.service`, after the 60-second sleep, the service exits itself instead of interrupting the already-running upgrade process. The service does report as "failed" when this happens, and I'm assuming that's not a problem.

The recovery partition updates seem to be working now; not sure what changed to fix that one. (The failures earlier were not causing the process to exit, so it wasn't the same as the notifier issue.)

jacobgkau (changes requested):
The unmasking timeout fix is not working because line 221 (which now has a timeout) is only hit if there's a failure. The timeout also needs to be applied on line 215, for when there's no failure during the rest of the upgrade.

jacobgkau:
@mmstick The timeout still isn't working. Maybe we can just manually attempt to delete the `/dev/null` symlinks from `/etc/systemd/system/` to unmask? Otherwise, this needs further debugging to see why it would hang when trying to unmask a non-existent service. The screen currently hangs on every upgrade, so this is easy to recreate in a VM.

jacobgkau:
Actually, running it again and modifying the script to have the extra messages again, I just got a successful reboot for the first time in a VM. The lemp13 was what failed earlier; I'm going to do some more testing.

jacobgkau:
After more testing, I'm not seeing the issue occur in a VM anymore, but I am still seeing it happen on physical hardware, including both lemp13 and addw2. On lemp13, I checked to make sure the `timeout` commands were there in the script before rebooting, but did not modify the script. On addw2, I added additional messages and confirmed that it's still getting stuck on the unmask command, even with the timeout. (After running the upgrade, it's impossible to recheck the script, since `pop-upgrade` currently gets removed and the script's not there anymore in 24.04.)

jacobgkau:
gdm3 and gnome-shell were still present after an upgrade on fc16e12. I pushed a commit to correct the name of the `gdm3` package (there is no `gdm` package, [it's `gdm3` for all suites](https://packages.ubuntu.com/noble/gdm3)), and also to remove everything starting with `gnome-shell-extension` instead of just the prefs (because `gnome-shell-extension-pop-x11gestures` was still installed, and that was what kept `gnome-shell` installed as a dependency).

jacobgkau:
GNOME removal's finally working with the latest changes. It wasn't running at all without the `-y` flag.

The manual unmasking seems to be working; the services are not masked upon reboot.

I need to do a test of the latest version on physical hardware, and also a test with additional packages added to make sure removing all of those old workarounds isn't going to cause problems.

jacobgkau:
This is the upgrade.log from the lemp13 after hard-rebooting from where it got stuck: [upgrade.log](https://github.com/user-attachments/files/17139613/upgrade.log)

It looks like the issue is related to `system76-driver`, which is why it didn't occur in the VM.

It could be because `gnome-shell-extension-system76-power` is now being removed. `system76-driver` depends on it: https://github.com/pop-os/system76-driver/blob/12b4d159995678617daa30dc96d81cba86eec646/debian/control#L29 

Unlike `gnome-shell-extension-pop-x11gestures`, `gnome-shell-extension-system76-power` does not depend on `gnome-shell` (even though it should), so keeping it installed wouldn't prevent `gnome-shell` from being removed. However, `gnome-shell-extension-system76-power`  is obsolete, so we want to be able to remove it, anyway. We need to make `system76-driver` no longer hard-depend on that package, possibly in a `_noble` branch (which currently isn't separated out).

mmstick:
There should be a new system76-driver package in noble now

jacobgkau:
`system76-driver` is not being removed now, but System76 hardware is still hanging on `Replacing GNOME...` 🙃

[upgrade.log](https://github.com/user-attachments/files/17157498/upgrade.log)

Not sure why it's doing this. When I forcibly reboot, I need to run `sudo dpkg --configure -a` before I can use apt, but it just picks up in the middle of where it was, starting with `Setting up policykit-1-gnome...`, as if nothing was wrong.

mmstick:
It seems be caused by `pop-wallpapers`, because removing it removes the provides for `ubuntu-wallpapers`, which causes a number of other packages to get removed alongside it.

mmstick:
Should be fixed by https://github.com/pop-os/cosmic-wallpapers/commit/8d1d23ab64c52ad57f09c1e5e79585cc97dd4eea

jacobgkau:
Still hanging on `Removing GNOME....` with that latest `cosmic-wallpapers` update. [upgrade.log](https://github.com/user-attachments/files/17171348/upgrade.log)

Comparing the two upgrade logs, it got stuck in exactly the same place, and it was the same set of GNOME-related packages being removed. I don't think `cosmic-wallpapers` was related to this issue.

mmstick:
Updated pop-desktop to have speech-dispatcher and orca as recommended dependencies, so they shouldn't get autoremoved now. These are similarly recommended for the ubuntu-desktop metapackage.

jacobgkau:
Alright, from a clean install, it successfully finishes upgrading and reboots on lemp13 now.

jacobgkau (approved):
I've performed a few more test upgrades, including one with Steam, WINE, kde-full, and the third-party PPAs for Spotify and qBittorrent, which succeeded.

This is looking stable enough for pre-release upgrades.

data/pop-upgrade.sh
```diff
@@ -239,29 +184,41 @@ attempt_upgrade () {
     if (upgrade || attempt_repair); then
         rm -rf  /system-update "$1"
 
+        # Set cosmic-greeter as the default display manager
+        rm -f /etc/systemd/system/display-manager.service
```
jacobgkau, line 188: I tried doing a 20.04 to 22.04 upgrade, since this is merging into `master` and there are no `master_*` branches in this repo.

GDM was not uninstalled, but the system booted to a TTY after the upgrade because this line removed GDM as the default. This part needs to be gated behind a version check, too (or we need to branch off `master_focal` before merging this).

jacobgkau (changes requested):

data/pop-upgrade.sh
jacobgkau: As a third alternative to the version check or separate branches, this block could alternatively check if COSMIC Greeter's installed before making it the default. If a version check's simpler, though, it should give the same end result as of right now.

jackpot51 (dismissed):

mmstick:
I'm guessing that it's asking for an answer. Need to pass the flags for unnattended upgrades.

mmstick:
It's probably this

```
  policykit-1-gnome
0 upgraded, 1 newly installed, 89 to remove and 1 not upgraded.
Need to get 22.4 kB of archives.
```

It shouldn't be trying to fetch packages while offline.

mmstick:
Should we add libpam-fprintd as a recommended dependency for noble?

jacobgkau:
> Should we add libpam-fprintd as a recommended dependency for noble?

GDM is what pulls it into Jammy. Would that be something we want to add as a recommends for COSMIC Greeter, or is it more generally usable by other components, too?

It does look like at least one person's using a fingerprint reader with COSMIC Greeter, looking through its issue tracker. I don't have a compatible fingerprint reader to check the situation with.

mmstick:
This may help: https://github.com/pop-os/cosmic-osd/commit/c6fda40a0aa42c07dc357ebed6ce935cee3269ae

jacobgkau:
The same scenario as before now succeeds instead of getting stuck. `policykit-1-gnome` is not installed after finishing the upgrade. The `cosmic-greeter` suggests is successfully keeping `libpam-fprintd` installed on a system where it was previously installed.

If I don't find anything else, this might be good to merge early next week.

jacobgkau (approved):
This is still working, not seeing any more issues.

jackpot51 (approved):

kmf:
Blog Mentions, and links to this issue.
> **Updating and Upgrading to COSMIC Alpha 3**
>To switch from a previous alpha version to COSMIC Alpha 3, check the Updates page in COSMIC ?>Store and make sure your version is up to date. Alpha testers on Pop!_OS 22.04 LTS can now 
[upgrade to 24.04 LTS](https://github.com/pop-os/upgrade/pull/330) using the developer flag.

Perhaps update the Blog entry with command, or add a comment here?

The Command?
`env S76_TEST=1 pop-upgrade release upgrade`

wie-was:
> So, the current complete command to upgrade is `pop-upgrade release upgrade -f`, as I stated over at [#335 (comment)](https://github.com/pop-os/upgrade/issues/335#issuecomment-2377213636).

Hypothetically, is there an easy way to downgrade back to 22.04?

mmstick:
Using a 22.04 image to refresh the system

hawko2600:
Do you need to accommodate for this?

Nov  5 22:20:54 pop-os kernel: [23370.487632] audit: type=1400 audit(1730805654.619:46): apparmor="DENIED" operation="open" class="file" profile="ubuntu_pro_esm_cache" name="/etc/pop-os/os-release" pid=34406 comm="python3" requested_mask="r" denied_mask="r" fsuid=0 ouid=0

Nov  5 22:20:54 pop-os python3[34406]: ["2024-11-05T22:20:54.620", "ERROR", "ubuntupro.lib.esm_cache", "main", 17, "Error updating the cache: [Errno 13] Permission denied: '/etc/os-release'", {}]

PR #331, opened 2024-06-03
i18n(ru): Translation update

FAlexei:


jacobgkau (approved):
The newly translated string shows up as expected in a user account set to Russian.

mmstick (approved):

PR #334, opened 2024-09-25
Update pop_upgrade_gtk.ftl (French)

roxfr:


jacobgkau (approved):
The "System and Application Updates" section header is now translated when viewing with a user account set to French.

PR #341, opened 2024-11-08
fix: Don't downgrade conflicting Ansible packages in the same command

### LINKED ISSUE #340: Fails to downgrade packages when Ansible is installed from PPA
As reported by @hawko2600 at https://github.com/pop-os/upgrade/pull/330#issuecomment-2461357517:

```
Nov 5 22:21:18 pop-os pop-upgrade[35075]: The following packages have unmet dependencies:
Nov 5 22:21:18 pop-os pop-upgrade[35075]: ansible-core : Breaks: ansible (< 4.6.0-1~) but 2.10.7+merged+base+2.10.8+
dfsg-1 is to be installed
Nov 5 22:21:18 pop-os pop-upgrade[35075]: Recommends: ansible (>= 4.6.0-1~) but 2.10.7+merged+base+2.
10.8+dfsg-1 is to be installed
Nov 5 22:21:18 pop-os pop-upgrade[35075]: E: Unable to correct problems, you have held broken packages.
Nov 5 22:21:18 pop-os pop-upgrade[28948]: [https://github.com/pop-os/upgrade/pull/33[1mINFO https://github.com/pop-os/upgrade/pull/33[0m] daemon/mod.rs#033[0m#033[1m:https://github.com/pop-os/upgrade/pull/33[0m#033[1;35m362#033[0m: upgrade result: Err(Downgrade(apt-get downgrade
Nov 5 22:21:18 pop-os pop-upgrade[28948]: Caused by:
Nov 5 22:21:18 pop-os pop-upgrade[28948]: status is unknown: exit status: 100))
Nov 5 22:21:18 pop-os pop-upgrade[35078]: Canceled hold on pop-upgrade.
```

I'm able to recreate this from a fresh VM installation by adding the Ansible PPA and installing `ansible` before attempting the upgrade.

jacobgkau:
This fixes #340 and should work with all corner cases I can think of regarding the Ansible packages. Ubuntu's weird Ansible packaging seems to only exist on 22.04 (20.04 didn't have an `ansible-core` package, so I'm guessing it was this way for a historical reason when the `ansible-core` package was first being introduced); for that reason, this case handling checks the version number and only applies for the 22.04 version (for upgrades to 24.04).

Testing (WIP):
- [x] When `ansible-core` and `ansible` were installed from the PPA, removes `ansible-core` and keeps `ansible` installed (fixes #340).
    - In 24.04, `ansible-core` is installed again as a dependency of `ansible`.
- [x] When only `ansible-core` is installed from the PPA (and `ansible` is not installed from anywhere), keeps `ansible-core` installed.
    - The PPA's `ansible` depends on `ansible-core`, so the reverse (having the PPA's `ansible` but no `ansible-core`) is not a possible configuration.
- [ ] ~~When `ansible-core` is installed from the PPA but `ansible` is installed from Ubuntu (via apt pin), removes `ansible-core`.~~
    - Not a possible configuration due to file conflicts.
- [x] When `ansible-core` is installed from Ubuntu (via apt pin) but `ansible` is installed from the PPA, removes `ansible-core`.
    - Works because `ansible` is still downgraded to the Ubuntu version, and because `ansible-core` isn't on the downgrade list, apt is able to automatically remove it.
    - In 24.04, `ansible-core` is installed again as a dependency of `ansible`.
- [x] When either `ansible-core` or `ansible` is installed from the Ubuntu repo (without the PPA added), keeps it installed.
    - [x] `ansible-core` only from Ubuntu
    - [x] `ansible` only from Ubuntu
        - In 24.04, `ansible-core` is installed as a dependency of `ansible`.
    - The 22.04 Ubuntu repo doesn't allow installing both at once.
- [x] Upgrade with neither package installed still succeeds.

In all of the cases where `ansible` was installed from any source, once on 24.04, Ubuntu's `ansible-core` is installed as a dependency of Ubuntu's `ansible`, because this weird Ubuntu packaging is only on 22.04. (In all cases where `ansible-core` only was installed from any source, once on 24.04, only `ansible-core` is installed.)


jacobgkau:
All of my testing is passing after including https://github.com/pop-os/apt-cmd/pull/9 in this fix.

mmstick (approved):

leviport (approved):

PR #339, opened 2024-11-05
fix: add Signed-By fields to apt sources

mmstick:

PR #343, opened 2024-12-31
Don't attempt to use `apt.pop-os.org/ubuntu` on aarch64

ids1024:
This mirror does not currently have aarch64 packages, so use `ports.ubuntu.com` again.

### DIFF
--- a/daemon/src/release/repos.rs
+++ b/daemon/src/release/repos.rs
@@ -341,16 +341,18 @@ fn ubuntu_uri() -> &'static str {
 }
 
 fn system_sources(release: &str) -> String {
+    let uri =
+        if cfg!(target_arch = "aarch64") { ubuntu_uri() } else { "apt.pop-os.org/ubuntu" };
     format!(
         r#"X-Repolib-Name: Pop_OS System Sources
 Enabled: yes
 Types: deb deb-src
-URIs: http://apt.pop-os.org/ubuntu
+URIs: http://{1}
 Suites: {0} {0}-security {0}-updates {0}-backports
 Components: main restricted universe multiverse
-X-Repolib-Default-Mirror: http://apt.pop-os.org/ubuntu
+X-Repolib-Default-Mirror: http://{1}
 "#,
-        release
+        release, uri
     )
 }
 


mmstick (dismissed):

ids1024:
Not sure why `noble` builds on this repository are failing now.

mmstick:
Same as https://github.com/pop-os/upgrade/pull/339

The GTK library isn't needed on Noble, so that could be removed there.

mmstick:
Fixed the noble builds by disabling the GTK library there.

leviport:
I gave this a quick whirl on the Pi 400, and the in-OS part of the upgrade succeeded, but it seems stuck after the reboot:
IMG_20250103_150422_129

I let it sit there for a while since the Pi 400 isn't the fastest rig in the world, but after 45m, I think it's pretty solidly stuck.

jacobgkau:
On the x86 side, it doesn't look like this causes any regressions. The URLs are still set properly (to our mirror) when they're updated from e.g. Jammy to Noble. The system successfully boots into Noble after the upgrade.

@leviport Does the apt sources file at least look good after it's updated on the Pi?

leviport:
Yeah the sources file looks fine. I'd imagine we don't need to hold this up because of Pi upgrade issues.

ids1024:
For reference, it did upgrade successfully on my Pinebook Pro. Though I just had it run overnight, so I don't know exactly how long it took.

debian/rules
```diff
@@ -3,6 +3,9 @@
 export VENDOR ?= 1
 CLEAN ?= 1
 
+VERSION=$(shell lsb_release -rs)
+ENABLE_GTK_LIB=$(shell dpkg --compare-versions 22.04 le $(VERSION) && echo 1 || echo 0)
```
ids1024, line 7: Shouldn't this be `$(VERSION) le 22.04` rather than `22.04 le $(VERSION)`?

ids1024:
>``` gnome-control-center: error while loading shared libraries: libpop-upgrade-gtk-ffi.so.1: cannot open shared object file: No such file or directory```

Maybe removing the `.install` files here isn't going to work.

I think that CI issue should be fixed by https://github.com/pop-os/at-spi2-core/pull/2. Then https://github.com/pop-os/upgrade/pull/343/commits/60f65d0a4ee088ba5af2c5e539e023881ebb184e isn't necessary, though that package isn't useful on noble. May need to create a branch to properly remove that on only noble.

ids1024:
I've removed 60f65d0a4ee088ba5af2c5e539e023881ebb184e from the PR for now, given the above mentioned issue(s).

With https://github.com/pop-os/at-spi2-core/pull/2 merged, this is building. So it probably should be good to merge now.

leviport:
Update: the Pi 400 upgrade worked! I just never saw the Plymouth screen, and the output on the screen (similar to https://github.com/pop-os/upgrade/pull/343#issuecomment-2569878235) looks a bit broken. But when I returned to check on it again, I found it sitting at the COSMIC DE login screen.

ids1024:
Yeah, I think it was similar on the Pinebook Pro. So there may be a bug to look into, but it does ultimately upgrade.

PR #346, opened 2025-02-28
fix: add exemption for sentinelagent package

### LINKED ISSUE #347: Fail to upgrade if SentinalOne endpoint protection is installed
Tried to run an upgrade but it consistently fails with `Release upgrade status: release upgrade aborted: conflicting and/or deprecated packages could not be removed` with the help of the COSMIC Epoch channel was able to narrow it down to 

here is my apt history:

[history.log](https://github.com/user-attachments/files/19033496/history.log)

I cannot manually remove it either need to first get a passphrase from our Corporate admin to temporarily remove it

mmstick:
There may be other safe packages that we could add here.

Closes #347

### DIFF
--- a/daemon/src/release/mod.rs
+++ b/daemon/src/release/mod.rs
@@ -584,7 +584,10 @@ async fn remove_conflicting_packages(logger: &dyn Fn(UpgradeEvent)) -> Result<()
     .map_err(ReleaseError::ConflictRemoval)?;
 
     // Add packages which have no remote to the conflict list
-    if let Ok(packages) = apt_cmd::apt::remoteless_packages().await {
+    if let Ok(mut packages) = apt_cmd::apt::remoteless_packages().await {
+        // Add exemptions for specific packages that we know to be safe.
+        packages.retain(|name| name != "sentinelagent");
+        // Add the packages that remain to our list of conflicting packages for removal.
         conflicting.extend_from_slice(&packages);
     }
 


jacobgkau:
Mattermost thread for background: https://chat.pop-os.org/pop-os/pl/oup58puaabr1xfcjhaqj47ob6h 

The `sentinelagent` package includes a pre-removal script that fails purposefully, because Sentinel prefers users use their official uninstallation script, which requires a password from the organization's admin. It is possible to forcefully uninstall by removing files and having dpkg ignore the pre-removal script, though. https://www.cybervigilance.uk/insights/sentinelone-uninstalling-agent 

Installing `sentinelagent` again after it's removed would require a .deb file and additional information from the organization's admin. https://jumpcloud.com/support/install-the-sentinelone-agent#linux 

Because of that, and because we don't want organizations to regard Pop!_OS as overriding their device management policies, we're exploring the possibility of allowing this package, which is orphaned (by design) and thus would normally be removed, to remain installed through an upgrade.

Besides this approach, the third alternative would be presenting a message to the user that they need to have their organization's admin remove `sentinelagent` for them before proceeding with the upgrade (essentially getting permission for the upgrade).

DW8Reaper:
@mmstick & @jacobgkau just to let you know I ran pop-upgrade successfully with this fix and didn't run into any issues. My SentinelOne agent also started up without issues so it seems to be running fine. 

Thanks for the quick help

jacobgkau (approved):
Thank you for testing, @DW8Reaper. We can't QA the `sentinelagent` package itself because it's only available to SentinelOne customers.

An upgrade with only the default Pop!_OS packages succeeded.
An upgrade with a different orphaned package (`davinci-resolve`) still removed that package (as expected) and succeeded.

wash2 (approved):

PR #348, opened 2025-03-01
Add Hungarian translations

therealmate:


jacobgkau (approved):
This translation is showing up as expected in a user account set to Hungarian.

A sanity check through Google Translate checks out.

wash2 (approved):

Issue #353, opened 2025-05-01
Keep gnome-online-accounts-gtk on upgrades


WatchMkr:
To avoid breaking online accounts when upgrading from 22.04 and allow users to remove online accounts that were added.

PR #354, opened 2025-05-13
feat: Keep gnome-online-accounts-gtk when removing GNOME

jacobgkau:
Fixes https://github.com/pop-os/upgrade/issues/353.

### DIFF
--- a/data/pop-upgrade.sh
+++ b/data/pop-upgrade.sh
@@ -201,7 +201,7 @@ attempt_upgrade () {
             message -i "Upgrade complete. Replacing GNOME..."
             apt-get remove --autoremove -y ~nlanguage-pack-gnome ~ngnome-user-docs gdm3 gnome-bluetooth gnome-calendar \
                 gnome-contacts gnome-online-miners gnome-orca gnome-shell ~ngnome-shell-extension gnome-themes-standard \
-                gnome-tweaks
+                gnome-tweaks gnome-control-center gnome-online-accounts-gtk+
         fi
 
         message -i "Upgrade complete. Updating initramfs for all kernels..."


jacobgkau:
This appears to be working. `gnome-online-accounts-gtk+` (with the `+` at the end) tells apt to keep that package installed. `gnome-control-center` now needs to be explicitly added to the removal list, since `gnome-online-accounts` otherwise keeps it as a recommends (it's not actually needed, as `gnome-online-accounts-gtk` is an alternative to that recommends).

jacobgkau:
Discovered this doesn't work if the internet is disconnected after downloading the upgrade but before the "offline" (rebooted) portion of the upgrade occurs. This is because `gnome-online-accounts-gtk` is a new package with 24.04 (it's not even available in the 22.04 repos), and while I assumed it was downloaded in advance as a dependency of `gnome-control-center` or `gnome-online-accounts`, it's actually not. The upgrade doesn't fail, but it's left without Online Accounts if there's no internet during the `Removing GNOME...` step of the offline upgrade. To fix this, just need to ensure that package is downloaded by pop-upgrade in advance.

jacobgkau:
Getting an error regarding the `apt install --print-uris` command with 4b0bdeb.

jacobgkau:
On trying this again today, it seems to be working in a fresh VM:

Screenshot_2025-06-03_13-20-45

After completing the upgrade, Online Accounts is installed, but GNOME Shell and GNOME Control Center are no longer installed, as expected. This works even if internet is removed after downloading and before rebooting to install the upgrade.

However, I still got the error again in the VM that gave that error last time:

Screenshot_2025-06-03_14-23-48

Attempting to replicate the issue manually (by updating the source lists to `noble` and running `sudo apt full-upgrade --print-uris` followed by `sudo apt --print-uris install gnome-online-accounts-gtk`), the package resolver's complaining about an unrelated `pop-desktop` package dependency that will be solved by the `full-upgrade` but would not be solved by the `install` command alone.

Screenshot_2025-06-03_14-33-01

If I change the `install` command to `download`, then it prints the URI successfully, since that one doesn't try to resolve the rest of the system's dependencies. I'll try to figure out how to have this usage of the `apt_fetch` function do that instead, while still having it get dependencies for its usage earlier in the process.

jacobgkau:
This seems to be working now! gnome-online-accounts-gtk is fetched at the end of pop-upgrade's download process:

Screenshot_2025-06-18_02-21-16

After disconnecting the VM's network adapter and rebooting into the offline portion of the upgrade, the upgrade completes as expected. On 24.04, gnome-online-accounts-gtk is present as expected and opens up successfully (also tested adding & removing a Google account successfully), while other GNOME components such as gnome-shell and gnome-control-center are no longer present.

We need to get https://github.com/pop-os/apt-cmd/pull/10 merged first now that we know that part's working (and make a release of that), then point pop-upgrade's Cargo.toml back to a normal release of the crate. Also want to do a little more testing to make sure the changes here in pop-upgrade don't have any hidden side effects.

jacobgkau:
Tested that this works when internet is disabled during the post-reboot portion of the upgrade, and still works when internet is enabled during that part, too. Also checked that automatic updates still work, that `pop-upgrade release repair` still works, and that core packages are still installed as part of the `release upgrade` process if they're missing beforehand.

leviport (approved):

PR #360, opened 2025-09-24
Switchable graphics default

### LINKED ISSUE #358: Reset system76-power graphics mode to the default the installer sets up
Usually this will be hybrid mode. This avoids hybrid laptop upgrades being left in nvidia mode and the GPU always being on (COSMIC no longer has or needs a GUI to switch between modes).

mmstick:
- Closes #358 
- Updates rust version to 1.89.0
- Fixed EOL unit test
- Removed a vulnerable crate identified by [cargo-audit](https://crates.io/crates/cargo-audit)
- Replaced many unmaintained crates also identified by `cargo-audit`
    - All that remains is is the gtk3 crates being unmaintained, which we'll soon drop


jacobgkau (approved):
- [X] On a test addw2 (Open Firmware), I first set `system76-power graphics nvidia` and rebooted (then confirmed I was in NVIDIA mode), then upgraded with this branch applied. After preparing the upgrade (but before rebooting), running `system76-power graphics` reported Hybrid, although the GNOME Shell extension still had the dot on NVIDIA mode (and did not state anything was pending a reboot). After rebooting into the offline portion of the upgrade and finishing the upgrade, once I was rebooted into 24.04, `system76-power graphics` reported that the mode is Hybrid.

- [X] I repeated the test on oryp5 (proprietary firmware), but starting in Integrated mode. It had the same behavior, ending up in Hybrid mode as expected.

I also did the following additional tests:

- [X] darp7 (physical Intel system) 22.04 with KDE installed: no issues
- [X] Relatively fresh VM 22.04 -> 24.04 upgrade: no issues
- [ ] Problem packages VM 20.04 -> 22.04 upgrade
    - This is a 20.04 VM with a wide range of intentionally problematic packages installed.
    - Failed in the offline portion of the upgrader on an NVIDIA and/or kernel issue (got stuck in a "Checking for package fixes..." loop for hours, between which it would configure NVIDIA, the kernel, or process triggers for initramfs-tools). This PR isn't touching the offline upgrade script, and I also tested the same upgrade with the version of pop-upgrade currently released for 20.04 (e21935f) and saw the same issue, so it's definitely not a regression with this PR.

PR #361, opened 2025-09-29
fix: reset login shell on upgrade to /bin/bash

mmstick:
Fix for missing login shells on systems upgrading with a login shell installed from a third party PPA or debian package, such as nushell.


jacobgkau:
Installed nushell from https://github.com/nushell/integrations?tab=readme-ov-file#install-nushell-for-ubuntudebian.

- Current stable (f437d90): Confirmed that an upgrade completed successfully, but attempting to log into COSMIC Epoch resulted in a `/usr/bin/nu: No such file or directory` message, followed by being sent back to COSMIC Greeter.
- This branch (7566247): The same upgrade completed successfully, and I was able to log in as expected afterwards. My shell in `/etc/passwd` is set to `/bin/bash`, which matches the default for Pop!_OS users. No other passwd lines appear to be affected (even `/bin/sync` for the `sync` user, which isn't covered by the `login_is_disabled` function). A `usermod` line appeared in journalctl for my own user only regarding the shell change.

Looks good so far.

jacobgkau (approved):
Tried one more test upgrade with three users-- the primary & an additional with Bash, and a third with Fish installed. When upgrading from the primary user, the secondary user with Fish had its shell changed back to Bash, while the two Bash users had corresponding `no changes` lines logged. After the upgrade, all three users could still be logged into, and while the Fish package was still installed (as expected), the Fish user defaulted to Bash on login.

Checking if each user's shell still exists and, if not, performing the reset at the end of the upgrade might be less disruptive than resetting all users' shells towards the beginning of the upgrade. But seeing as third-party sources are removed earlier in the process than that, doing it here doesn't seem wrong. This is accomplishing its goal and not causing any regressions with upgrades.

PR #369, opened 2025-10-28
fix: Install standard Wacom packages over Surface variants

### LINKED ISSUE #364: Failing to remove conflicting libwacom Microsoft Surface packages
Version: 22.04
Kernel: Linux pop-os 6.16.9-surface-1 SMP PREEMPT_DYNAMIC
Device: Surface Pro 6 

I have a fresh PopOS install with some adjustment to the kernel (linux-surface) so I can use the touchscreen and some applications fron `apt` and Pop Store. When I tried to upgrade to Pop 24.04 to use COSMIC, it appeared to be like this: 

```
Release upgrade status: release upgrade aborted: conflicting and/or deprecated packages could not be removed
```
I have try apt update, upgrade and full-upgrade, where it said that it has nothing to update.
Having read `journalctl` and `dmesg`, there wasn't anything way too unusual. 

`journalctl -o cat -fu pop-upgrade` output (the "important" part): 

```
[INFO ] daemon/mod.rs:874: release upgrade: removing deprecated and/or conflicting packages
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 libinput10 : Depends: libwacom9 (>= 2.0.0)
              Depends: libinput-bin (>= 1.20.0-1ubuntu0.3)
 libqt5svg5 : Depends: libqt5widgets5 (>= 5.15.3+dfsg~) but it is not going to be installed
 pop-desktop : PreDepends: pop-de-gnome but it is not going to be installed
               Recommends: touchegg but it is not going to be installed
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
[INFO ] daemon/mod.rs:362: upgrade result: Err(ConflictRemoval(conflict removal
Caused by:
    status is unknown: exit status: 100))
Canceled hold on pop-upgrade.
[INFO ] daemon/mod.rs:366: setting upgrade state
```

I suspect this is caused by my custom kernel for the surface laptops. Nonetheless, can you please help me check it out? Thanks a bunch.

jacobgkau:
Apt fails to remove libwacom-common-surface and libwacom9-surface, which need to be removed because they're orphaned. We need to manually install the standard packages to work around this.

Fixes #364.

### DIFF
--- a/daemon/src/release/mod.rs
+++ b/daemon/src/release/mod.rs
@@ -152,7 +152,8 @@ pub enum UpgradeEvent {
     Failure = 13,
     AptFilesLocked = 14,
     RemovingConflicts = 15,
-    Simulating = 16,
+    RemovingWacomConflicts = 16,
+    Simulating = 17,
 }
 
 impl From<UpgradeEvent> for &'static str {
@@ -174,6 +175,7 @@ impl From<UpgradeEvent> for &'static str {
                 "ensuring that system-critical packages are installed"
             }
             UpgradeEvent::RemovingConflicts => "removing deprecated and/or conflicting packages",
+            UpgradeEvent::RemovingWacomConflicts => "replacing Surface-tailored Wacom packages with standard ones",
             UpgradeEvent::Success => "new release is ready to install",
             UpgradeEvent::SuccessLive => "new release was successfully installed",
             UpgradeEvent::UpdatingPackageLists => "updating package lists",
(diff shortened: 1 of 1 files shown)


jacobgkau:
@mmstick Could you take a look at this?

mmstick (approved):

leviport (approved):

PR #378, opened 2026-01-02
fix: ensure systemd-resolved is manually installed before upgrading

mmstick:
Fixes DNS issues on upgrade if systemd-resolved happens to be auto-removed


leviport (approved):

jacobgkau:
https://github.com/pop-os/desktop/blob/master_jammy/debian/control required `systemd-resolvconf | resolvconf`. https://github.com/pop-os/desktop/blob/master_noble/debian/control just recommends `systemd-resolved`.

It looks like `systemd-resolved` didn't exist on 22.04, and `systemd-resolved` and `systemd-resolvconf` conflict with each other on 24.04 (so only one can be installed at a time). (`systemd-resolvconf` was dropped from Ubuntu for 24.04, but is still being provided by our systemd package.)

Is the 24.04 control file set up correctly, or should we have a hard depends on `systemd-resolved | systemd-resolvconf | resolvconf` or something to replace what we had for 22.04?

Issue #379, opened 2026-01-07
Upgrade to 24.04 flow


maria-komarova:
The flow to upgrade from 22.04 to 24.04 ([mockups in Figma](https://www.figma.com/design/oyWySlcr0Re02kaiZbEG44/Upgrade?node-id=41-295&t=qA2lkMxOwloA7rPE-4)):

<img width="5233" height="2250" alt="Image" src="https://github.com/user-attachments/assets/259c10d6-f98f-4933-8136-9510c566848f" />

PR #380, opened 2026-01-15
feat: enable 24.04 upgrades for 22.04

mmstick:
Currently testing. It would be ideal to merge this as `master_jammy` and then allow `master` to target >=24.04. Then we can drop the GTK library entirely on 24.04 and make changes that are required by pop-upgrade integrations into cosmic-settings.


jacobgkau (changes requested):
When running through this for the first time in a virtual machine, I saw the first upgrade dialog open with a lot of empty space at the bottom:

Screenshot_2026-01-15_13-04-37

Let's also try to better match the text to what's shown in the Figma file linked in https://github.com/pop-os/upgrade/issues/379:

Screenshot_2026-01-15_13-06-23

In particular, the first line shouldn't be there, and the paragraph doesn't need to be a bullet point. Are we able to have hyperlinks in this dialog?

The second dialog also looks identical to the first at this point, where they should be different dialogs per the mockup.

jacobgkau (dismissed):
The good news is that this is generally working now!

Both of the upgrade dialogs still have a lot of empty space beneath them when opened:

Image

Image

I checked a physical machine to confirm this isn't just a quirk with my VM, but it happened there, too:

Image

So that will need to be fixed.

jacobgkau:
This is what the mockup looks like for the final dialog:

Screenshot_2026-02-10_13-05-36

This is what the dialog currently looks like on a laptop on battery power (the first sentence doesn't show up if it's plugged in when the dialog's opened):

Screenshot from 2026-02-10 14-53-12

@pop-os/ux Is this close enough, or should we add the line break and remove the bold to better match the mockup?

maria-komarova:
Bold is fine as is. The line break would help since there is an icon on the left. I don't know if we can make the dialog a bit narrower but that might help too.

jackpot51 (approved):

leviport (approved):

PR #385, opened 2026-02-07
style: fix and enforce clippy linting

danieleades:
adds clippy linting.

- move clippy config to the workspace config
- address all existing clippy errors
- add clippy CI job

marked as a draft since this is stacked on #383 and #384 and these should be merged first

## Note on Clippy version

this is currently using the pinned toolchain (`rust-toolchain.toml`) as the clippy version. Pinning clippy to a static version prevents new lints causing breakages in PRs which are unrelated to the changes in those PRs. That's the behaviour we want- but it's not strictly required to pin to *this* version- it could be a newer version of clippy (since clippy respects the `rust-version` MSRV).

i've elected not to do that here, because if you pinned clippy to a recent nightly version, you'd need an automated way to regularly bump that pinned version. [Renovate](https://docs.renovatebot.com/modules/manager/regex/) can do that- but this project isn't using Renovate.

Something to consider in future.

### DIFF
--- a/daemon/src/daemon/mod.rs
+++ b/daemon/src/daemon/mod.rs
@@ -312,7 +312,7 @@ impl Daemon {
                             info!("upgrading packages");
                             let _ = crate::release::package_upgrade(|event| {
                                 let _ = dbus_tx.send(SignalEvent::Upgrade(event));
-                            });
+                            }).await;
 
                             info!("packages upgraded");
                         }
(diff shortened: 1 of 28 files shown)


daemon/src/daemon/mod.rs
```diff
@@ -306,7 +312,7 @@ impl Daemon {
                             info!("upgrading packages");
                             let _ = crate::release::package_upgrade(|event| {
                                 let _ = dbus_tx.send(SignalEvent::Upgrade(event));
-                            });
+                            }).await;
```
danieleades: this was a bug- without awaiting the future, `package_upgrade` never ran

danieleades:
@mmstick does it make sense to bundle the formatting changes with this PR?

Issue #370, opened 2025-10-31
Upgrade fails if `dotnet-host` and `dotnet-sdk-x.0` installed from Microsoft repo


brianary:
When I try to upgrade to 24.04 on my Thelio Mira Premium that I ordered almost exactly a year ago, it fails:

```
$ sudo pop-upgrade release upgrade -f                                                                    
checking if pop-upgrade requires an update
Current Release: 22.04
Upgrading to: 24.04
New version available: 20
Recovery event: fetching recovery files
Recovery upgrade status: recovery partition refueled and ready to go
Event: updating package lists
Event: upgrading packages for the current release
Event: fetching updated packages for the current release
Release upgrade status: release upgrade aborted: failed to downgrade packages
```

I've included the _/var/log/syslog_ and _/var/log/apt/history.log_ files in _[log.zip](https://github.com/user-attachments/files/23277501/log.zip)_.

I tried booting from the 24.04 USB that I used to successfully install PopOS! on my Framework 13 laptop, but it failed:

<img width="950" height="700" alt="Image" src="https://github.com/user-attachments/assets/3d196de3-27a4-4508-a02c-625fb52da5be" />

Secure boot appears to be disabled.

I've also included _installer.log_ in _[log.zip](https://github.com/user-attachments/files/23277501/log.zip)_.

lincolnthalles:
I went through the same thing and worked it out.

While pop-upgrade disables third-party repositories (PPA), it doesn't remove the applications that were installed by them, and it doesn't give a helpful error.

In your case, the first blocker is dotnet. You can remove it with `sudo apt remove ~ndotnet`. If anything else blocks the upgrade, you can check with `journalctl | grep downgrading`.

Note that pop-upgrade will nuke any manually installed `deb` packages, and you'll have to reinstall them later. Flatpaks and applications from the official repositories will survive the upgrade just fine.

jacobgkau:
> While pop-upgrade disables third-party repositories (PPA), it doesn't remove the applications that were installed by them, and it doesn't give a helpful error.

It should remove the applications that were installed by them, because one of the next steps is removing orphaned packages, as you noted. (The exception is if the package is also available from the built-in sources, in which case we downgrade to that version.) I can try to see if I can replicate an issue with dotnet from https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-install?tabs=dotnet9&pivots=os-linux-ubuntu-2204#ubuntu-2204.

[...]

PR #386, opened 2026-02-10
fix: Remove dotnet-host if dotnet-sdk-8.0 is also installed

jacobgkau:
dotnet-sdk-8.0 depends on dotnet-host in Microsoft's repository, but dotnet-sdk-8.0 conflicts with dotnet-host in the Ubuntu repository. If they're both installed, the latter must be removed before the former's downgraded.

Fixes https://github.com/pop-os/upgrade/issues/370. Based on what I did for https://github.com/pop-os/upgrade/pull/341.

### DIFF
--- a/daemon/src/release/mod.rs
+++ b/daemon/src/release/mod.rs
@@ -562,6 +562,27 @@ async fn downgrade_packages() -> Result<(), ReleaseError> {
             }
             info!("ansible is not installed, so ansible-core will be downgraded");
         }
+        
+        // dotnet-sdk-8.0 depends on dotnet-host in Microsoft's repository,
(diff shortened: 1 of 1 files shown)


jacobgkau:
Confirmed this works to fix #370:

Screenshot_2026-02-09_20-50-07

The upgrade was successful, and `dotnet-sdk-8.0` was still installed after the upgrade was complete (along with `dotnet-host-8.0`, which provides `dotnet-host`).

Also confirmed that if `dotnet-host` is installed from the Microsoft repo but `dotnet-sdk-8.0` isn't, `dotnet-host` is not removed but is instead downgraded normally:

Screenshot_2026-02-09_21-32-31

Again, the upgrade was successful, with `dotnet-host` still installed after it was complete.

mmstick (approved):

PR #388, opened 2026-02-20
fix: download recovery ISO from generic channel in 24.04

mmstick:
Preserves fetching from the `intel` channel on 22.04


jackpot51 (approved):

jacobgkau (commented):
This is working across 22.04 and 24.04.

`master` is currently built for 20.04 (Focal) as well, with no separate `master_focal` branch. This branch had its 20.04 build fail. Is that expected? If so, we should probably go ahead and fork `master_focal`, if only so we don't accidentally release further updates to it. (Otherwise, if we intend to keep Focal on the same branch, then the "not 22.04" check probably needs to check for 20.04 as well-- that's what I was trying to test when I noticed the Focal build had failed.)

PR #390, opened 2026-03-02
Improvement: make script POSIX sh compatible

LinuxBoy-96:
This change makes upgrade.sh compatible with POSIX /bin/sh by removing bash-specific constructs. It has been tested with shellcheck, checkbashisms, and dash. No functional changes are intended. The goal is to improve portability for future use and integration in systems that may differ from the current environment.

Since this does not introduce new functionality and only improves compatibility, I am submitting it directly as a small refactor.

### DIFF
--- a/data/do-release-upgrade
+++ b/data/do-release-upgrade
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 
-echo "\`do-release-upgrade\` is not supported on Pop!_OS."
-echo "Use \`pop-upgrade release upgrade\` instead."
-echo ""
-echo "For more information, visit: https://support.system76.com/articles/upgrade-pop"
-
+printf '%s\n' \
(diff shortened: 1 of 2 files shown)


LinuxBoy-96:
> Is there anywhere other than Pop!_OS that would be trying to use the pop-upgrade utility?
I agree this utility is Pop!_OS-specific today, and that Bash is currently guaranteed on Pop!_OS via pop-container-runtime.

My comment about portability was more about keeping the script strictly POSIX when it doesn’t need any Bash-specific features. It makes the script easier to reuse and audit, and it keeps the door open for downstream forks (e.g., COSMIC-based systems) that may not want to rely on Bash by default.

Personally, I’m also experimenting with a minimal system design that avoids Bash as a hard dependency where possible, so this was a small preparatory step in that direction.

Issue #297, opened 2022-05-14
pop-upgrade 100% CPU usage


chase:
Running Pop_OS! 22.04.

I have automatic updates disabled, automatic install disabled, and update notifications set to Weekly.

I noticed that my fans started going full speed and htop showed that `pop-upgrade daemon` was at 100% CPU.
The process tree showed that there was an apt-get instance that was defunct.
I let it sit for a few hours before using `kill -9` to stop it.

---

Another symptom that may be related, running `apt update` shows that it is unable to reach `apt.pop-os.org`

Visiting apt.pop-os.org in Firefox does not seem to be an issue.

GeodeCotin:
Same issue here!

AnishDe12020:
It started recently for me too (22.04), the fan noise in unbearable, hope it is fixed soon!

Janvanoorschot:
Same here (22.04). 8 cores running full speed for more then 15 minutes. Its doing some SSL encrypted traffic with an amazonaws host. Upgrade... really?

Janvanoorschot:
Killed it with a -9 and it reported:
```
Aug 15 10:25:12 mouse systemd[1]: pop-upgrade.service: Consumed 45min 27.651s CPU time.
```

miaoerwu:
@mmstick @jackpot51 
I have the same problem. pop-upgrade daemon makes a cpu run 100% and does not stop.
This causes several problems (that I have found so far) as follows.
1. the laptop fan runs at high speed to dissipate heat, making a loud noise
2, the battery power drops rapidly, originally can use about 10 hours, now can only use about 2 and a half hours
3, need to wait for a long time to exit the process when shutting down, or forced shutdown
4. the os upgrade & recovery page in settings keeps loading without any result

Janvanoorschot:
I managed to trace it back (IMHO) to me activating Settings/OS Upgrade& Recovery/OS Recovery/Update Recovery Partition.

miaoerwu:
> I managed to trace it back (IMHO) to me activating Settings/OS Upgrade& Recovery/OS Recovery/Update Recovery Partition.

Yes, the pop-upgrade daemon process comes up after a while after each visit here

miaoerwu:
> @mmstick @jackpot51 I have the same problem. pop-upgrade daemon makes a cpu run 100% and does not stop. This causes several problems (that I have found so far) as follows.
> 
> 1. the laptop fan runs at high speed to dissipate heat, making a loud noise
[...]

PR #396, opened 2026-04-23
Fix: Remove Infinite Retry on Metadata Fetch To Prevent 100% CPU Usage

helios2k6:
This PR attempts to remove an infinite loop caused by a race between the network interface watcher and a network request. I suspect that this race is the root-cause behind [100% CPU issue](https://github.com/pop-os/upgrade/issues/297).

Errors that occur during the metadata fetch will simply be propagated up and handled.

### DIFF
--- a/daemon/src/misc.rs
+++ b/daemon/src/misc.rs
@@ -83,4 +83,3 @@ pub fn uid_min_max() -> anyhow::Result<(u32, u32)> {
             Ok((min, max))
         })
 }
-
--- a/daemon/src/misc.rs
+++ b/daemon/src/misc.rs
@@ -1,12 +1,11 @@
 use anyhow::Context;
-use std::{fs, future::Future, io, path::Path, time::Duration};
+use std::{fs, io, path::Path, time::Duration};
 use tokio::fs::{copy, File};
 
 pub fn http_client() -> Result<reqwest::Client, reqwest::Error> {
     reqwest::ClientBuilder::new()
-        // Error if the HTTP header response, or the time to fetch a chunk of a body, exceeds 30 seconds.
+        .connect_timeout(Duration::from_secs(30))
         .read_timeout(Duration::from_secs(30))
-        // Limit to following 10 redirects before failing.
         .redirect(reqwest::redirect::Policy::limited(10))
         .build()
 }
@@ -85,23 +84,3 @@ pub fn uid_min_max() -> anyhow::Result<(u32, u32)> {
         })
 }
 
-pub async fn network_reconnect<Fun, Fut, Res>(func: Fun) -> Res
-where
-    Fun: Fn() -> Fut,
-    Fut: Future<Output = Res>,
-{
-    loop {
-        let request = func();
-        let changed = async_fetcher::iface::watch_change();
-
-        futures::pin_mut!(request);
-        futures::pin_mut!(changed);
-
-        use futures::future::Either;
-
-        match futures::future::select(request, changed).await {
-            Either::Left((result, _)) => break result,
-            Either::Right(_) => tokio::time::sleep(Duration::from_secs(3)).await,
-        }
-    }
-}
--- a/daemon/src/release/repos.rs
+++ b/daemon/src/release/repos.rs
@@ -152,8 +152,7 @@ pub async fn is_old_release(codename: &str) -> bool {
     let url = &["http://old-releases.ubuntu.com/ubuntu/dists/", codename, "/Release"].concat();
 
     if let Ok(client) = crate::misc::http_client() {
-        let request = || async { client.head(url).send().await };
-        if let Ok(resp) = crate::misc::network_reconnect(request).await {
+        if let Ok(resp) = client.head(url).send().await {
             return resp.status().is_success();
         }
     }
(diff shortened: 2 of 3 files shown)


mmstick (approved):

jacobktm (approved):
<details>
<summary>Regression testing passed:</summary>

## Checklists

Tasks for a tester to verify when approving a patch.

### Daemon

All frontends interact with the daemon, and therefore testing a frontend will also test the daemon. However, there are ways that you can alter the system environment to make it hostile towards the success of a release upgrade. These are scenarios that the daemon should be able to recover from.

- [ ] The recovery partition is missing from `/etc/fstab`.
    - daemon does not recover from modifications to /etc/fstab. Not a regression.
- [ ] The EFI partition is missing from `/etc/fstab`.
    - daemon does not recover from modifications to /etc/fstab. Not a regression.
- [ ] UUID mount points are switched to PartUUID.
    - daemon does not recover from modifications to /etc/fstab. Not a regression.
- [x] Missing Pop repositories are re-added.
- [x] `pop-desktop` is installed if it is not installed.
- [x] Interactively allows the user to handle incompatible PPAs.
- [x] Handles the `us.archives` -> `old-releases` transition for EOL'd upgrades.

### CLI

Features which can be tested from the command line interface. Each command gives detailed output which is not seen in the GTK frontend. When testing, report any wordings or colors that could be improved to give the user a better experience when using the command line.

- [x] `pop-upgrade recovery default-boot from-release` boots into the recovery partition on the next boot.
- [x] `pop-upgrade recovery upgrade` upgrades the recovery partition.
- [x] `pop-upgrade release check` reports the current, next, and release availability.
- [x] `pop-upgrade release refresh` boots into the recovery partition in refresh mode.
- [ ] `pop-upgrade release repair` fixes a number of common system issues that may prevent an upgrade.
    - This command doesn't seem to actually do anything from the CLI. Not a regression.
- [x] `pop-upgrade release update` is equivalent to `apt update && apt full-upgrade`, but much faster.
- [x] `pop-upgrade release upgrade` updates the current release, and prepares for a release upgrade.
    - [x] `pop-upgrade release upgrade -f` forces an upgrade, even if the next release is a development branch.
- [x] `pop-upgrade status` returns a string describing the status of the daemon (ie: `inactive`).

### GTK

This frontend shares much of the same client code as the CLI, so by testing this, you are also testing much of the same behaviors seen in the CLI. This is what the majority of users will interact with.

- [x] On a system without a new release, it should report that no new releases are available.
- [x] Selecting "Refresh OS" should boot into the recovery partition and skip to the "Refresh OS" view.
- [x] "Refresh OS" should not be an available option on a system without a recovery partition.
- [x] Selecting "Start Upgrade" should present a dialog with a changelog detailing key items in the new release.
    - [x] Selecting "Start Upgrade" on an upgradeable release with a recovery partition.
        - [x] The recovery partition should be upgraded in this scenario.
    - [x] Selecting "Start Upgrade" on an upgradeable release without a recovery partition.
        - [x] On success, the reboot will proceed into offline updates mode.
- [x] Once "Start Upgrade" is clicked, a progress bar appears to display the current step and its progress.
- [x] Options for dismissing desktop notifications must be present on LTS releases.
- [x] Dismissing an update on LTS shall prevent desktop notifications from appearing until the next release.

### Notifications

Similar to Firmware Manager, a systemd user timer will occasionally ask the upgrade daemon if a new release is available.

- [x] If a release is available, a desktop notification will display with non-actionable text declaring that an update is available.
- [x] On clicking the notification, this will open GNOME Settings to allow the user to either dismiss or perform the upgrade.
- [x] Non-LTS releases should not be allowed to dismiss an upgrade.
- [x] Double-check all text and icons in the notification, including version numbers.
- [x] The "Upgrade" button should still appear when notifications are dismissed.

### Plymouth

Critical for offline upgrades with systemd, the Plymouth theme presents information from the upgrade script executed at init to the user in a way that does not overwhelm them with information. If Plymouth is disabled, the user will simply see a black screen with a lot of scrolling text.

- [x] When performing an offline upgrade with systemd, our Plymouth theme will actively show progress as it occurs.
- [x] The Plymouth screen will show the Pop! logo, but it should not pulsate.
- [x] Make sure all text is shown on the screen and can be read. Text should not spill off the edges of the screen.

### Systemd Upgrades

When the recovery partition does not exist, systemd's `offline-updates` service is used to perform a release upgrade. The following are guidelines for testing this functionality.

- [x] Test on a variety of possible configurations in the wild:
    - [x] With, and without, a recovery partition
    - [ ] Custom kernels
        - Upgrade fails if a custom built kernel was not installed from a .deb package. Not a regression.
    - [x] Multi-boot installations
    - [x] LVM on LUKS (encrypted)
    - [x] Upgradeable and incomatible PPAs
- [x] NVIDIA testing:
    - [x] Old cards that require the legacy driver.
    - [x] Graphics switching, including changing graphics modes, not rebooting, and initiating the upgrade.
    - [x] Upgrades with our CUDA packages.
- [x] Release upgrades should always upgrade to the latest-available release.
- [x] Verify that the recovery partition is accessible after the upgrade.
- [x] Test upgrades on systems that have fully up-to-date packages, and ones that have updates available.
- [x] Verify that the Plymouth screen looks correct:
    - The Pop! logo should appear, but should not pulsate.
        - Update progress should be shown as a percentage. 
        - All text should be readable on the screen, and should not spill off the edges.
- [x] The OS name in the EFI boot menu should be updated to reflect the new release.

</details>

PR #399, opened 2026-06-18
Prevent busy-spin on dbus connection error in main event loop - Possibly addresses #297

orensbruli:
Fix the busy-spin loop when connection.process() returns an error.

### DIFF
--- a/daemon/src/daemon/mod.rs
+++ b/daemon/src/daemon/mod.rs
@@ -797,7 +797,11 @@ impl Daemon {
         let mut shutdown_triggered = false;
 
         loop {
-            let _ = connection.process(std::time::Duration::from_millis(500));
+            if connection.process(std::time::Duration::from_millis(500)).is_err() {
+                warn!("dbus connection error, retrying in 5s");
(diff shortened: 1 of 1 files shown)


mmstick:
Is this AI-generated? The description doesn't match the contents at all and the change blocks the executor and can cause an infinite loop. Listing the number of insertions and deletions also isn't useful.

orensbruli:
Yeah @mmstick, I have been trying to fix this for some time, and after a long session with the AI it looks like it's now working for me.

I'm closing the PR because Rust is not my cup of tea, and I would not be able to honestly discuss with you the right approach for this :pray: 

Having said that, do you think that it could be that a persistent error in this call
```
let _ = connection.process(std::time::Duration::from_millis(500));
```
could create a busy-spin that would consume 100% of the CPU?

Thank you, and sorry for taking your time on this.

mmstick:
The busy loop that was reported was networking-related rather than DBus. It was reported to be fixed last month by f2903ce2631c18d721530ee263aff6e31c5951b1

PR #403, opened 2026-09-09
Enable 26.04 development upgrades

mmstick:


jacobgkau (changes requested):
After adding this branch, `pop-upgrade release upgrade` still does nothing, but adding `-f` now allows an upgrade to 26.04 to be attempted. However, an upgrade in a relatively fresh 24.04 VM failed. We don't need to do extensive testing just to enable development upgrades, but it seems like it'd be a good idea for at least a basic case of a fresh install to work before we enable this.

The first error I saw was about the recovery partition failing to upgrade:

Screenshot_2026-09-10_10-33-44

Looking at the logs, I think this one's actually just because the ISO's not published yet. The error message is misleading about that cause, though, so we might want to look into improving the message shown in that situation.

The more critical error came later:

Screenshot_2026-09-10_10-43-44

Screenshot_2026-09-10_10-44-32

Any idea why coreutils and sudo would have a problem at this stage of the upgrade? (I assume it might have to do with Ubuntu moving to the Rust versions of these?)

mmstick:
Last change didn't fix it. Will update when I get a system upgraded.

mmstick:
The apt-get simulate argument can give false positives in this scenario so I've removed it. Upgraded all of my systems with pop-upgrade to validate and they each successfully completed without any issues.

jacobgkau (commented):
I'm still having a problem when upgrading my fresh 24.04 VirtualBox VM. About a third of the way into the offline portion of the upgrade, I'm dropping from Plymouth (or the upgrade output) to what appears to be a non-responsive decryption prompt. It took a few attempts of trying to track this down to realize that it wasn't actually rebooting, but instead just dropping to this screen for some reason while the upgrade's still running in the background.

I took a screen recording after having pressed Esc to show the log instead of Plymouth. This was the last frame before it dropped to the false decryption prompt:

mpv-shot0002

`plymouth-theme-ubuntu-text` and `fonts-ubuntu` are the last two packages that were set up before the problem occurs. The next thing I see is this (again, same screen whether I'd previously pressed Esc or not):

Screenshot_2026-09-14_13-04-15

The first couple of times this happened, I tried typing the decryption passphrase, saw nothing happen, and then did a hard reboot (which isn't a good thing to do in the middle of an upgrade). I'd imagine we'd see users do similar things if this was widespread. This last time, I noticed the hard drive activity light was still blinking, so I waited a while, and the upgrade eventually did finish as expected.

Any idea what could be breaking the view of progress during that offline portion of the upgrade?

jacobgkau (changes requested):
I attempted an upgrade on physical hardware (oryp12) today, and the same issue (with losing all indications of progress about a third of the way through) occurred. So this isn't just a VM quirk, and will definitely be a blocker for the upgrade.

Additionally, the cryptsetup prompt after the initial reboot was in text mode instead of showing the Plymouth theme. This hasn't happened for previous release upgrades.

mmstick:
So it's just encrypted installs having this problem? I do recall seeing a package stall the progress of the upgrade for a minute or two but the unencrypted upgrades I did finished successfully. I'll do some encrypted upgrades today.

jacobgkau:
> So it's just encrypted installs having this problem?

I tried a non-encrypted VM (on edcd1e7), and it had basically the same problem-- about a third of the way through, the screen simply went blank for the rest of the upgrade, instead of showing the cryptsetup text prompt like the encrypted installs did. I'll wait to try 95e6b4b or a later commit if you're still working on it.

mmstick:
I'm currently checking. Maybe one of the commits I made after I did the upgrades have caused that.

mmstick:
LUKS encrypted installs are now upgrading. The LUKS config for dracut was not being applied. Now the decryption screen is using the plymouth GUI and I didn't encounter any issues with the upgrade stalling or breaking.

I also added a fix to distinst to prevent the unsquashfs command from failing due to OOM from excessive cache usage. Which I was encountering when doing an encrypted install in a VM with 16 GB RAM.

jacobgkau (changes requested):
The progress bar is now showing up through the entirety of the offline portion of the upgrade, as expected.

After upgrading, `fwupd-unsigned` and `pop-server` are both still on their 24.04 versions:

Screenshot_2026-09-17_12-39-10

An `apt full-upgrade` replaces `libcanberra0t64` with `libcanberra0`, but does not upgrade those two packages. Trying to `apt install` them seems to pull in `dracut` and remove `initramfs-tools`:

Screenshot_2026-09-17_12-42-47

This seems like something that should be taken care of during the offline upgrade script?

mmstick:
Yes, a046acd should be upgrading pop-server with dracut once the upgrade is complete.

mmstick:
Followup to confirm that the last commit fixed the packages that were not upgraded.

jacobgkau (approved):
An upgrade of a fresh 24.04 install now shows the progress as expected and results in a completely up-to-date 26.04 system with all of the expected packages. The upgrade is still gated behind the `-f` / `--force-next` flag, so this is fine to add for testing and development purposes at this point. Any further problems we discover can be fixed before the upgrade is marked as stable.

jacobgkau:
I would normally want to give this a quick test with a 22.04 -> 24.04 upgrade since `master` currently includes Jammy, but this branch is only built for Noble and Resolute. We might want to sanity-test it in master before we release the update to Jammy, or branch `master_jammy` off if we want to avoid any inadvertent regressions for the 22.04 -> 24.04 upgrade. Some users are definitely still on 22.04 and waiting out future COSMIC updates.

mmstick:
We will branch `master_jammy`

jackpot51 (approved):

